PR #6, opened 2021-03-30
:bug: fixing a bug with saturated pixels and tpf meta data

christinahedges:
There are a few bugs in the new implementation

1) The saturated pixel mask masks the wrong entries
2) If there are "none" values in the tpfs for the KEPMAG keyword, the machine breaks


christinahedges:
See [#7]

PR #10, opened 2021-06-01
[Do Not Merge] Sketch of potential refactor

christinahedges:
I think this is a rough sketch of what we need in PSF machine, benefitting from `kepler-apertures`. Porting the capabilities we like from `kepler-apertures` to `psfmachine` will mean we only have to maintain one package. `kepler-apertures` is in a great shape for us to do this merge. 

Some key thoughts:
* We probably need an FFI class
* We need a good ability to save/load PRF models in the `TPFMachine` class
* We probably don't need a superstamp class, we can probably make a proceedure that uses the FFI machine for that.
* We have to watch out for where we can be reasonably generic and allow TESS data, and where we need to stop users adding tess data (for now)
* I haven't included any diagnostic plots, but I think those would be great.
* I really want the functionality to be able to do something like

```python
tpm = psf.TPFMachine.from_tpfs(tpf)
tpm.get_lightcurves()
```

which would populate a `tpm.lcs` attribute (or something similar) with `lk.LightCurve` objects, each of which contains the PRF photometry and the SAP photometry. 

Crucially, I don't want it to fit anything, I just want to load the stored weight files and apply them really fast.

### DIFF
--- a/src/psfmachine/ffi.py
+++ b/src/psfmachine/ffi.py
@@ -1,0 +1,118 @@
+"""Classes for working with FFI data"""
+from .utils import get_gaia_sources
+from .machine import Machine
+
+
+def FFIMachine(Machine):
+    """Subclass of Machine for working with FFI data"""
+
+    # Probably don't need a very new init function over Machine.
+    def __init__(self, channel=channel, quarter=quarter, **kwargs):
+        super().__init__(**kwargs)
+        self.channel = channel
+        self.quarter = quarter
+
+    def __repr__(self):
+        return f"FFIMachine (N sources, N times, N pixels): {self.shape}"
+
+    @staticmethod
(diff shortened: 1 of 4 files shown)


jorgemarpa:
This is great. Here some of my thoughts:

- An FFI class will make sense, it will inherit from `machine` but it will adapt some things that are necessary to work with really large sparse matrices for efficiency, this is already included in `kepler-apertures`.
- `TPFmachine` will need to load PRF models from files, and create apertures for: 
       - A given source
       - All sources in a TPF
- One method for `TPFmachine` that does things under the hood: 
       - get PRF models
       - create apertures
       - compute aperture photometry
       - create `lightkurve.LightCurve` objects with the light curves
 
I'll work these ideas in different pull requests and branches so they will be easier to assess.

PR #12, opened 2021-06-05
Load and save PRF shape model 

jorgemarpa:
I added two new methods:

- `load_shape_model()`: to load the parameters that define a PRF model. `machine` ends up in the same state as when running `build_shape_model()`
- `save_shape_model()`: to save the parameters that define a PRF model

Also included small improvements:
- Avoid circular import of PACKAGEDIR variable
- Include pandas
- Added the option to load or fit model to `fit_lightcurves(load_shape_model=True)` so we can get PSF light curves from the PRF model computed from the FFIs.
- Change from `tpf.header` to `tpf.get_header()` to avoid deprication warnings. 

The new functionalities work ok, no execution errors. But there's one BIG problem. The PRF models I created from the FFIs are in pixels! and `machine` works in arcseconds. I hardcode the conversion `r [pix] = r [arcsec] /4` when computing `self.mean_model` but the result doesn't look right (see next image). 😞  Will look more into it. 

PRF model for quarter 12 channel 4:
image


jorgemarpa:
Fixed the scale problem, by adding a factor of 4 to `rmin, rmax, cut_r` so the design matrix is created in arcsecond scale.
Also removed background pixels from the mean_model
Now I notice that the PRF models from the FFIs are flipped, here's a case where is horizontally flipped  (channel 4): 

using PRF from FFI
image

using PRF from TPFs
image

Edit: 
I think the flip issue is also because the PRF models are in pixel space. When `psfmachine` does the pix to radec conversion the coordinate grid flips.

src/psfmachine/tpf.py
```diff
@@ -283,6 +294,130 @@ def plot_tpf(self, tdx):
...
+
+        # Recreate uncontaminated mask
+        self._get_uncontaminated_pixel_mask()
+
+        # Recreate mean model!
+        self._get_mean_model()
+
+        if plot:
+            return self.plot_shape_model()
+        return
```
christinahedges, line 364: From line 330-386: I think this little loop (or something similar) is also undertaken in `build_shape_model`. I think it's removing "background" pixels. We should pull this out as it's own helper method, so that we're not repeating ourselves.

src/psfmachine/tpf.py
```diff
@@ -283,6 +294,130 @@ def plot_tpf(self, tdx):
...
+        # check if file exists
+        if not os.path.isfile(input):
+            raise FileNotFoundError("No PSF files: ", input)
+
+        try:
+            tab = pd.read_csv(input, index_col=0, header=[0, 1])
+            self.n_r_knots = int(tab.loc[channel, (str(quarter), "n_r_knots")])
+            self.n_phi_knots = int(tab.loc[channel, (str(quarter), "n_phi_knots")])
+            # *4 to convert from pixel to arcseconds
+            self.rmin = int(tab.loc[channel, (str(quarter), "rmin")] * 4)
```
christinahedges: This *4 is only valid when we're using a kepler-apertures format...we need to be careful here.

src/psfmachine/tpf.py
```diff
@@ -283,6 +294,130 @@ def plot_tpf(self, tdx):
...
+        output : str, None
+            Output file name. If None, one will be generated.
+        """
+        # asign a file name
+        if output is None:
+            output = "%s/data/shape_model_ch%02i_q%02i.csv" % (
+                PACKAGEDIR,
+                self.tpf_meta["channel"][0],
+                self.tpf_meta["quarter"][0],
+            )
```
christinahedges, line 379: Is it possible to add a mission keyword here? I don't know if TPFMachine carries that meta data around but I think it might

I think we should make a separate directory for TPFMachine and future FFIMachine

*These ones should be automatically loaded when we use `load_shape_model`*
`data/PRFMachine/kepler_q01_ch01_shape_model.csv`
*These ones are more risky, they depend on how many TPFs were used.*
`data/TPFMachine/kepler_q01_ch01_shape_model.csv`

We may even prefer to have `save` automatically dump the CSV file in the current working directory, rather than storing them in the package.


Update: it should never go to the package directory, always working directory, we can put in the right package place.

src/psfmachine/tpf.py
```diff
@@ -283,6 +294,130 @@ def plot_tpf(self, tdx):
...
+            [self.n_r_knots, self.n_phi_knots, self.rmin, self.rmax]
+            + self.psf_w.tolist()
+        )
+        df_dict = {
+            self.tpf_meta["quarter"][0]: pd.DataFrame(
+                np.atleast_2d(arr_to_save),
+                index=[self.tpf_meta["channel"][0]],
+                columns=["n_r_knots", "n_phi_knots", "rmin", "rmax"]
+                + ["w%02i" % i for i in range(1, 1 + len(self.psf_w))],
+            )
```
christinahedges: We should probably save some comment meta data about e.g. the spline order, and maybe the keywords that made the file so that it's reproducible?

christinahedges (changes requested):
This is GREAT. Small PRs are the way forward. Having a drop in function that replaces `build_shape_model` is exactly what we want.

We'll get this stuff ironed out with the shapes, it'd be really nice to be able to reuse them. We should be careful since PSFMachine made weight files will differ from kepler-aperture made files........might need some error catches for this.

I made a few small comments but this is looking in great shape. I'll checkout your branch and try it out.

src/psfmachine/tpf.py
jorgemarpa: that makes sense, we should put that in a separate function.

jorgemarpa: implemented in commit:
85ccd6461f95714826a9f64472cc0433942780d6

src/psfmachine/tpf.py
jorgemarpa: implemented here:
11d13f2b5269e39799ae72b5636abbce67a1783d

I will assume that PRF models from `kepler-apertures` will be named `ffi_prf_models_v*.csv`

src/psfmachine/machine.py
```diff
@@ -851,6 +851,14 @@ def build_shape_model(self, plot=False, flux_cut_off=1):
 
         # We then build the same design matrix for all pixels with flux
         self._get_mean_model()
+        # remove background pixels and recreate mean model
+        self._remove_background_data(flux_cut_off=flux_cut_off)
```
christinahedges: `_remove_background_data` --> `_remove_background_pixels`?

jorgemarpa:
I change the scope of this PR to save/load PRF models computed with PSFmachine only. Later I will include the functionality for PRF models from FFIs.

I decided to use FITS file to save the models, which makes it easy to store metadata. I included some important kewords in the metadata (example at the end).
The loading function checks that the provided file matches the data in `machine` object. Then load the necessary attributes to build the `mean_model`.

**Unit Test**
The test checks that  `psf_w`, `mean_model`, and other variables match the original state (when fitting a model) vs when loading a model from disk. 
I'm using a set of 20 TPFs for the above. I have to download (during testing) a new set of test TPF because those for testing (`test/data/tpf_*`) are not enough to fit a PRF model (singular matrix error). I suggest we update this set of TPFs to the one I'm using.


```
TTYPE1  = 'psf_w   '                                                            
TFORM1  = 'E       '                                                            
OBJECT  = 'PRF_shape'          / PRF shape parameters                           
DATATYPE= 'TPF stack'          / Type of data used to fit shape model           
ORIGIN  = 'PSFmachine.TPFMachine' / Software of origin                          
VERSION = '1.0.0   '           / Software version                               
MISSION = 'Kepler  '           / Mission name                                   
QUARTER =                   12 / Quarter of observations                        
CHANNEL =                    4 / Channel output                                 
N_RKNOTS=                   10 / Number of knots for spline basis in radial axis
N_PKNOTS=                   15 / Number of knots for spline basis in angle axis 
RMIN    =                    1 / Minimum value for knot spacing                 
RMAX    =                   16 / Maximum value for knot spacing                 
CUT_R   =                    6 / Radial distance to remove angle dependency in d
SPLN_DEG=                    3 / Degree of the spline basis                     
INC_INTE=                    T / Include or not interception for spline basis
```

src/psfmachine/machine.py
jorgemarpa: 👍🏾

src/psfmachine/machine.py
```diff
@@ -148,6 +148,7 @@ def __init__(
         self.time_radius = time_radius
         self.rmin = rmin
         self.rmax = rmax
+        self.cut_r = 6
```
christinahedges: Hard coding `self.cut_r` here, it should be a kwarg

src/psfmachine/machine.py
```diff
@@ -849,6 +851,14 @@ def build_shape_model(self, plot=False, flux_cut_off=1):
...
         # We then build the same design matrix for all pixels with flux
         self._get_mean_model()
+        # remove background pixels and recreate mean model
+        self._remove_background_pixels(flux_cut_off=flux_cut_off)
+
+        if plot:
+            return self.plot_shape_model()
+        return
+
+    def _remove_background_pixels(self, flux_cut_off=1):
```
christinahedges: Need docstring for what '1' is here. 1 count, 1%...? (I know that I wrote this 🤦‍♀️ )

christinahedges: JMP: This function updates the source mask, not the `psf_w`

CH: In that case, it needs a new name

christinahedges: Whole method needs a docstring

src/psfmachine/tpf.py
```diff
@@ -85,7 +92,9 @@ def __init__(
     def __repr__(self):
         return f"TPFMachine (N sources, N times, N pixels): {self.shape}"
 
-    def fit_lightcurves(self, plot=False, fit_va=True, iter_negative=True):
+    def fit_lightcurves(
+        self, plot=False, fit_va=True, iter_negative=True, load_shape_model=True
```
christinahedges: For now, let's have default False, because it won't work right now

src/psfmachine/tpf.py
```diff
@@ -283,6 +295,129 @@ def plot_tpf(self, tdx):
...
+            "Number of knots for spline basis in angle axis",
+        )
+        table.header["rmin"] = (self.rmin, "Minimum value for knot spacing")
+        table.header["rmax"] = (self.rmax, "Maximum value for knot spacing")
+        table.header["cut_r"] = (
+            self.cut_r,
+            "Radial distance to remove angle dependency",
+        )
+        table.header["spln_deg"] = (3, "Degree of the spline basis")
+        table.header["inc_inte"] = (
```
christinahedges: This is not a tuneable parameter for PSFmachine, we don't need to save it.

src/psfmachine/tpf.py
```diff
@@ -283,6 +295,129 @@ def plot_tpf(self, tdx):
...
+            self.n_phi_knots,
+            "Number of knots for spline basis in angle axis",
+        )
+        table.header["rmin"] = (self.rmin, "Minimum value for knot spacing")
+        table.header["rmax"] = (self.rmax, "Maximum value for knot spacing")
+        table.header["cut_r"] = (
+            self.cut_r,
+            "Radial distance to remove angle dependency",
+        )
+        table.header["spln_deg"] = (3, "Degree of the spline basis")
```
christinahedges, line 411: add a comment that spline degree is hard coded

christinahedges (approved):

christinahedges:
This is a great PR, looks like there are working tests and carefully thought out functions! Great work @jorgemarpa!

PR #13, opened 2021-06-08
tiled query

jorgemarpa:
This function divides the sky area into (5, 5)  cells. Then it finds the `ra`, `dec`, and `radius` that cover each cell and queries Gaia with those parameters. Later, concatenate all results and returns all unique sources in the original sky area.

### DIFF
--- a/src/psfmachine/utils.py
+++ b/src/psfmachine/utils.py
@@ -1,6 +1,7 @@
 """ Collection of utility functions"""
 
 import numpy as np
+import pandas as pd
 import functools
 
 from scipy import sparse
@@ -87,6 +88,77 @@ def get_gaia_sources(ras, decs, rads, magnitude_limit=18, epoch=2020, dr=2):
     else:
         raise ValueError("Please pass a valid data release")
     return gd.data.to_pandas()
+
+
+def do_tiled_query(
+    ra_grid, dec_grid, ngrid=(5, 5), magnitude_limit=18, epoch=2020, dr=3
+):
+    """
+    Find the centers and radius of the query when the sky area is large. This function
+    divides the data into `ngrid` tiles and compute the ra, dec coordinates for each
+    tile as well as its radius.
+
+    Parameters
+    ----------
+    ra_grid : numpy.ndarray
+        Data array with values of Right Ascension.
+    dec_grid : numpy.ndarray
+        Data array with values of Declination.
+    ngrid : tuple
+        Tuple with number of bins in each axis. Default is (5, 5).
+    magnitude_limit : int
+        Limiting magnitued for query
+    epoch : float
+        Year of the observation (Julian year) used for proper motion correction.
+    dr : int
+        Gaia Data Release to be used, DR2 or EDR3.
+
+    Returns
+    -------
+    sources : pandas.DatFrame
+        Pandas DatFrame with number of result sources (rows) and Gaia columns
+    """
+    # find edges of the bins
+    ra_edges = np.histogram_bin_edges(ra_grid, ngrid[0])
+    dec_edges = np.histogram_bin_edges(dec_grid, ngrid[1])
+    sources = []
+    # iterate over 2d bins
+    for ix in range(1, len(ra_edges)):
+        for jd in range(1, len(dec_edges)):
+            # check if image data fall in the bin
+            _in = (
+                (ra_edges[ix - 1] <= ra_grid)
+                & (ra_grid <= ra_edges[ix])
+                & (dec_edges[jd - 1] <= dec_grid)
+                & (dec_grid <= dec_edges[jd])
+            )
+            if not _in.any():
+                continue
+            # get the center coord of the query and radius
+            ra_in = ra_grid[_in]
+            dec_in = dec_grid[_in]
+            ra_q = ra_in.mean()
+            dec_q = dec_in.mean()
+            rad_q = np.hypot(ra_in - ra_q, dec_in - dec_q).max() + 10 / 3600
+            # print(
+            #     "Will do small queries query with this (ra, dec, radius, epoch): ",
+            #     ra_q,
+            #     dec_q,
+            #     rad_q,
+            #     epoch,
+            # )
+            result = get_gaia_sources(
+                tuple([ra_q]),
+                tuple([dec_q]),
+                tuple([rad_q]),
+                magnitude_limit=magnitude_limit,
+                epoch=epoch,
+                dr=dr,
+            )
+            sources.append(result)
+    #  concat results and remove duplicated sources
+    sources = pd.concat(sources, axis=0).drop_duplicates(subset=["designation"])
+    return sources
 
 
 def _make_A_polar(phi, r, cut_r=6, rmin=1, rmax=18, n_r_knots=12, n_phi_knots=15):
--- a/src/psfmachine/utils.py
+++ b/src/psfmachine/utils.py
@@ -135,15 +135,18 @@ def do_tiled_query(ra, dec, ngrid=(5, 5), magnitude_limit=18, epoch=2020, dr=3):
             )
             if not _in.any():
                 continue
-            # get the center coord of the query and radius 6th decimal (10 miliarcsec)
-            # to avoid not catching get_gaia_sources() due to floating point error.
+            # get the center coord of the query and radius to 7th decimal precision
+            # (3 miliarcsec) to avoid not catching get_gaia_sources() due to
+            # floating point error.
             ra_in = ra[_in]
             dec_in = dec[_in]
             # we use 50th percentile to get the centers and avoid 360-0 boundary
-            ra_q = np.round(np.percentile(ra_in, 50), decimals=6)
-            dec_q = np.round(np.percentile(dec_in, 50), decimals=6)
+            ra_q = np.round(np.percentile(ra_in, 50), decimals=7)
+            dec_q = np.round(np.percentile(dec_in, 50), decimals=7)
+            # +10/3600 to add a 10 arcsec to search radius, this is to get sources
+            # off sensor up to 10" distance from sensor border.
             rad_q = np.round(
-                np.hypot(ra_in - ra_q, dec_in - dec_q).max() + 10 / 3600, decimals=6
+                np.hypot(ra_in - ra_q, dec_in - dec_q).max() + 10 / 3600, decimals=7
             )
             # query gaia with ra, dec, rad, epoch
             result = get_gaia_sources(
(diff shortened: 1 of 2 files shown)


src/psfmachine/utils.py
```diff
@@ -89,6 +90,77 @@ def get_gaia_sources(ras, decs, rads, magnitude_limit=18, epoch=2020, dr=2):
     return gd.data.to_pandas()
 
 
+def do_tiled_query(
+    ra_grid, dec_grid, ngrid=(5, 5), magnitude_limit=18, epoch=2020, dr=3
```
christinahedges: `ra_grid` --> `ra`?

src/psfmachine/utils.py
```diff
@@ -89,6 +90,77 @@ def get_gaia_sources(ras, decs, rads, magnitude_limit=18, epoch=2020, dr=2):
...
+):
+    """
+    Find the centers and radius of the query when the sky area is large. This function
+    divides the data into `ngrid` tiles and compute the ra, dec coordinates for each
+    tile as well as its radius.
+
+    Parameters
+    ----------
+    ra_grid : numpy.ndarray
+        Data array with values of Right Ascension.
```
christinahedges: what shape is this array?

src/psfmachine/utils.py
```diff
@@ -89,6 +90,77 @@ def get_gaia_sources(ras, decs, rads, magnitude_limit=18, epoch=2020, dr=2):
...
+    -------
+    sources : pandas.DatFrame
+        Pandas DatFrame with number of result sources (rows) and Gaia columns
+    """
+    # find edges of the bins
+    ra_edges = np.histogram_bin_edges(ra_grid, ngrid[0])
+    dec_edges = np.histogram_bin_edges(dec_grid, ngrid[1])
+    sources = []
+    # iterate over 2d bins
+    for ix in range(1, len(ra_edges)):
```
christinahedges: `ix`, `jd`? 
`idx`, `jdx`?

src/psfmachine/utils.py
```diff
@@ -89,6 +90,77 @@ def get_gaia_sources(ras, decs, rads, magnitude_limit=18, epoch=2020, dr=2):
...
+                & (ra_grid <= ra_edges[ix])
+                & (dec_edges[jd - 1] <= dec_grid)
+                & (dec_grid <= dec_edges[jd])
+            )
+            if not _in.any():
+                continue
+            # get the center coord of the query and radius
+            ra_in = ra_grid[_in]
+            dec_in = dec_grid[_in]
+            ra_q = ra_in.mean()
```
christinahedges: this will break when it wraps at 360

src/psfmachine/utils.py
```diff
@@ -89,6 +90,77 @@ def get_gaia_sources(ras, decs, rads, magnitude_limit=18, epoch=2020, dr=2):
...
+            dec_q = dec_in.mean()
+            rad_q = np.hypot(ra_in - ra_q, dec_in - dec_q).max() + 10 / 3600
```
christinahedges: probably similar problem here

christinahedges: add a test for an ra patch that goes over 0-360

christinahedges (changes requested):
This looks great! Add in a few unit tests, and then we can merge this!

- [x] Unit test of a small patch
- [x] Unit test of RA/Dec patch over 0-360 degree boundary  
- [x] Unit test for FFI like data
- [x] Unit test for stack of TPF like data

The other thing is should we pretty much always use a tiled query? If yes make sure that this is the default?

Maybe there should be some helper function that decides if your data needs tiling, and how many tiles to use?

src/psfmachine/utils.py
jorgemarpa: done

src/psfmachine/utils.py
jorgemarpa: docstring added

src/psfmachine/utils.py
jorgemarpa: done

src/psfmachine/utils.py
jorgemarpa: fixed

src/psfmachine/utils.py
jorgemarpa: unit tests added

jorgemarpa:
I included the requested changes and also added the necessary unit tests for the function:
* FFI like data
* TPF stack: check the tiled query contain all sources from the simple query.
* Query sky area near 360->0 deg.

tests/test_tpfmachine.py
```diff
@@ -86,3 +89,73 @@ def test_parse_TPFs():
 def test_from_TPFs():
     c = TPFMachine.from_TPFs(tpfs)
     assert isinstance(c, Machine)
+
+
```
christinahedges, line 94:
```suggestion
@pytest.mark.remote_data
```

tests/test_tpfmachine.py
```diff
@@ -86,3 +89,73 @@ def test_parse_TPFs():
...
+        ra,
+        dec,
+        ngrid=(2, 2),
+        magnitude_limit=18,
+        dr=3,
+        epoch=epoch,
+    )
+    assert isinstance(sources_tiled, pd.DataFrame)
+    assert set(["ra", "dec", "phot_g_mean_mag"]).issubset(sources_tiled.columns)
+    assert sources_tiled.shape == (83, 11)
```
christinahedges: We might change these test files, or something might go awry. 
```suggestion
    assert sources_tiled.shape == sources_org.shape
```

tests/test_tpfmachine.py
```diff
@@ -86,3 +89,73 @@ def test_parse_TPFs():
...
+    # check that the tiled query contain all sources from the non-tiled query.
+    # tiled query is always bigger that the other.
+    assert set(sources_org.designation).issubset(sources_tiled.designation)
+
+    # test for FFI like data ch=4 q=12
+    # create a ra, dec grid similar to the FFI data. Here for convinience, I use the
+    # wcs solution from the TPF to conv the pixel grid to radec.
+    row = np.arange(1024)
```
christinahedges: this is a HUGE query. It will massively slow down tests. I would make it much smaller (maybe 100x100 tops)

tests/test_tpfmachine.py
```diff
@@ -86,3 +89,73 @@ def test_parse_TPFs():
...
+            0.0,
+        )
+        .T
+    )
+
+    ffi_sources = do_tiled_query(
+        ra,
+        dec,
+        ngrid=(5, 5),
+        magnitude_limit=18,
```
christinahedges: Go to a less faint magnitude limit, to make this query faster.

tests/test_tpfmachine.py
```diff
@@ -86,3 +89,73 @@ def test_parse_TPFs():
...
+        dr=3,
+        epoch=epoch,
+    )
+    assert ffi_sources.shape == (15822, 11)
```
christinahedges: I think hard coding this shape is probably unavoidable.

tests/test_tpfmachine.py
```diff
@@ -86,3 +89,73 @@ def test_parse_TPFs():
...
+
+    # Unit test for 360->0 deg boundary. we use a smaller sky patch now.
+    row = np.arange(100)
```
christinahedges, line 131: Smaller query to make this run faster. 10x10 is totally reasonable.

src/psfmachine/utils.py
```diff
@@ -89,6 +90,76 @@ def get_gaia_sources(ras, decs, rads, magnitude_limit=18, epoch=2020, dr=2):
...
+                & (dec <= dec_edges[jdx])
+            )
+            if not _in.any():
+                continue
+            # get the center coord of the query and radius 6th decimal (10 miliarcsec)
+            # to avoid not catching get_gaia_sources() due to floating point error.
+            ra_in = ra[_in]
+            dec_in = dec[_in]
+            # we use 50th percentile to get the centers and avoid 360-0 boundary
+            ra_q = np.round(np.percentile(ra_in, 50), decimals=6)
```
christinahedges: If the decimal part is vital here, let's make it 7 instead of 6, to ensure that it's fractions of an arcsecond...

src/psfmachine/utils.py
```diff
@@ -89,6 +90,76 @@ def get_gaia_sources(ras, decs, rads, magnitude_limit=18, epoch=2020, dr=2):
...
+            dec_q = np.round(np.percentile(dec_in, 50), decimals=6)
+            rad_q = np.round(
+                np.hypot(ra_in - ra_q, dec_in - dec_q).max() + 10 / 3600, decimals=6
```
christinahedges: Why is there a `+10/3600` here? are you hard coding a radius value? I think you may want to be careful about hard coding that, since we may want to use this in a very different regime in the future...

Something to think about. If nothing else, add a comment explaining what you're hard coding and why it's important.

christinahedges (changes requested):
Hey @jorgemarpa this is looking fantastic. I've left a few comments, once this is done this looks good to merge!

src/psfmachine/utils.py
jorgemarpa: Is to increase the search radius by 10". This is to allow for sources off sensor (up to 10").

tests/test_tpfmachine.py
jorgemarpa: the tiled query will always return more sources compared to the original query function for a TPF stack. The tiled query is not optimized for sparse sky patches. This is why I am testing that the tiled query **contains** the results of the original query. They won't have the same shape.

tests/test_tpfmachine.py
jorgemarpa: Agree it's a big query. I was testing that can manage FFI-like data. 
If you still wanna avoid this, we can combine this test with the one below and only do one 100x100 pix at the 0->360 boundary. That will cover both cases.

tests/test_tpfmachine.py
jorgemarpa: Combine this test with the FFI-like data in one 100x100 pix query at the RA boundary

tests/test_tpfmachine.py
jorgemarpa: If we do `sources_tiled.shape == sources_org.shape` then we have to run `_clean_source_list()` for both results first

jorgemarpa (commented):
Added the comparison of both results after cleaning the source list. Remove hardcoded variable shape.

christinahedges (approved):

christinahedges:
LGTM, great job @jorgemarpa!

PR #17, opened 2021-06-12
Machine goes full sparse arrays

jorgemarpa:
Upgrading `machine` to be fully functional with `scipy.sparse` arrays. 
This took some refactoring of how to create `dra`, `ddec`, `r`, and `phi` attributes and how they are used.

The method `_get_source_mask()` has the most changes depending on if using sparse or non-sparse arrays.

The current implementation works ok. All `machine` methods work, but is not optimal:
- Creating sparse arrays can be improved by replacing the for-loop with a tiled version
- Evaluate `<` for sparse arrays is inefficient. I implemented a better option but could be improved when the evaluation is done against an array of values (not a fixed value).

**TODO**

- [x] Unit tests
- [x] evaluate when non-sparse breaks and sparse is inefficient according to N sources and N pixels
- [x] profile functions


jorgemarpa:
Profiling results are that sparse arrays are efficient when `n_sources * N-pixels > 1e7`.
Test both implementations (numpy and sparse) for Kepler's TPF stacks, cluster mask, and FFIs.


image

src/psfmachine/machine.py
```diff
@@ -749,7 +857,7 @@ def plot_time_model(self):
         cbar.set_label("Normalized Flux")
         return fig
 
-    def build_shape_model(self, plot=False, flux_cut_off=1):
+    def build_shape_model(self, plot=False, flux_cut_off=1, **kwargs):
```
jorgemarpa, line 860: add `kwargs` so we can pass custom parameters to `_get_source_mask` if necessary.

src/psfmachine/machine.py
```diff
@@ -956,13 +1067,13 @@ def plot_shape_model(self, radius=20):
...
-            self.uncontaminated_source_mask.multiply(self.dra.value),
-            self.uncontaminated_source_mask.multiply(self.ddec.value),
+            self.uncontaminated_source_mask.multiply(self.dra),
+            self.uncontaminated_source_mask.multiply(self.ddec),
         )
         dx = dx.data * u.deg.to(u.arcsecond)
         dy = dy.data * u.deg.to(u.arcsecond)
 
-        fig, ax = plt.subplots(2, 2, figsize=(8, 6.5))
+        fig, ax = plt.subplots(2, 2, figsize=(9.5, 6.5))
```
jorgemarpa, line 1076: improve figure by avoiding y label overlap

src/psfmachine/machine.py
```diff
@@ -1028,7 +1139,8 @@ def plot_shape_model(self, radius=20):
             xlim=(-radius, radius),
             ylim=(-radius, radius),
         )
-
+        ax[0, 0].set_aspect("equal", adjustable="box")
```
jorgemarpa, line 1142: keeps ddec vs dra plot with a square aspect ratio

src/psfmachine/machine.py
```diff
@@ -1028,7 +1139,8 @@ def plot_shape_model(self, radius=20):
...
+        ax[1, 0].set_aspect("equal", adjustable="box")
```
jorgemarpa, line 1143: keeps ddec vs dra plot with a square aspect ratio

src/psfmachine/machine.py
christinahedges: Is it better to have a specific keyword? That way we can have it in a docstring.

src/psfmachine/machine.py
```diff
@@ -1132,3 +1245,38 @@ def fit_model(self, fit_va=False):
             self.werrs[:, nodata] *= np.nan
 
         return
+
+
+def sparse_lessthan(arr, limit):
+    """
+    Compute less than operation on sparse array by evaluating only non-zero values
+    and reconstructing the sparse array.
+
```
christinahedges, line 1264: Can you add a quick comment explaining why we need a function for this, and how this function works in a few words?

christinahedges: might be a comment in the other file, you could move that here

christinahedges (approved):
Hi @jorgemarpa this looks fantastic, I can't find any serious problems here. If you move that comment around I'll get this merged!!

jorgemarpa:
Awesome! I added a few words explaining how that sparse function works and why is needed.

christinahedges:
Thanks for this @jorgemarpa it's great work! I'm merging!

PR #19, opened 2021-06-16
Optional focus mask

### LINKED ISSUE #18: TPFMachine always masks for focus change
`TPFMachine` always masks focus change, and that's potentially not good behavior. We should have a flag to turn this off. 

https://github.com/SSDataLab/psfmachine/blob/7665dd52c5ee2541dcf3e3d79c64df2c1426b474/src/psfmachine/tpf.py#L69

jorgemarpa:
Masking the time array for focus changes is optional now.  I include the decision of creating or not a focus mask within the `.from_TPFs()` method. This way `TPFMachine` can still accept a `focus_mask` argument (boolean array or None) to be combined with the `time_mask`.

Also remove duplicated code, `focus_mask` was been calculated inside `_parse_TPFs` and `TPFMachine.__init__'.

This PR closes #18

### DIFF
--- a/src/psfmachine/tpf.py
+++ b/src/psfmachine/tpf.py
@@ -78,9 +78,9 @@ class TPFMachine(Machine):
         # )
         if time_mask is None and focus_mask is None:
             self.time_mask = np.ones(len(self.time), bool)
-        elif time_mask is None and not focus_mask is None:
+        elif time_mask is None and focus_mask is not None:
             self.time_mask = focus_mask
-        elif not time_mask is None and focus_mask is None:
+        elif time_mask is not None and focus_mask is None:
             self.time_mask = time_mask
         else:
             self.time_mask = time_mask & focus_mask
--- a/src/psfmachine/tpf.py
+++ b/src/psfmachine/tpf.py
@@ -66,21 +66,24 @@ class TPFMachine(Machine):
         self.tpfs = tpfs
 
         # Cut out 1.5 days after every data gap
-        dt = np.hstack([10, np.diff(time)])
-        focus_mask = ~np.in1d(
-            np.arange(len(time)),
-            np.hstack(
-                [
-                    np.arange(t, t + int(1.5 / np.median(dt)))
-                    for t in np.where(dt > (np.median(dt) * 5))[0]
-                ]
-            ),
-        )
-        if time_mask is None:
+        # dt = np.hstack([10, np.diff(time)])
+        # focus_mask = ~np.in1d(
+        #     np.arange(len(time)),
+        #     np.hstack(
+        #         [
+        #             np.arange(t, t + int(1.5 / np.median(dt)))
+        #             for t in np.where(dt > (np.median(dt) * 5))[0]
+        #         ]
+        #     ),
+        # )
+        if time_mask is None and focus_mask is None:
+            self.time_mask = np.ones(len(self.time), bool)
+        elif time_mask is None and not focus_mask is None:
             self.time_mask = focus_mask
+        elif not time_mask is None and focus_mask is None:
+            self.time_mask = time_mask
         else:
             self.time_mask = time_mask & focus_mask
-
         self.pix2obs = pix2obs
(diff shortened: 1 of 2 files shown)


src/psfmachine/tpf.py
```diff
@@ -65,22 +65,15 @@ def __init__(
...
-        dt = np.hstack([10, np.diff(time)])
-        focus_mask = ~np.in1d(
-            np.arange(len(time)),
-            np.hstack(
-                [
-                    np.arange(t, t + int(1.5 / np.median(dt)))
-                    for t in np.where(dt > (np.median(dt) * 5))[0]
-                ]
-            ),
-        )
```
jorgemarpa, line 78: duplicated coded, also seen inside `_parse_TPFs`

src/psfmachine/tpf.py
```diff
@@ -420,6 +413,7 @@ def from_TPFs(
         magnitude_limit=18,
         dr=2,
         time_mask=None,
+        do_focus_mask=True,
```
jorgemarpa: new kwarg for `from_TPFs`

tests/test_tpfmachine.py
```diff
@@ -154,14 +154,8 @@ def test_do_tiled_query():
 
 @pytest.mark.remote_data
 def test_load_save_shape_model():
-    # load sufficient TPFs to build a shape model
-    # the 10 TPFs in ./data/tpf_test* are not enough to fit a model, singular matrix
-    # error.
-    tpfs_k16 = lk.search_targetpixelfile(
-        "Kepler-16", mission="Kepler", quarter=12, radius=200, limit=10, cadence="long"
-    ).download_all(quality_bitmask=None)
```
jorgemarpa, line 162: this is not needed now, we can use the original test TPFs

src/psfmachine/tpf.py
christinahedges: can we have something like
`apply_focus_mask`?

src/psfmachine/tpf.py
```diff
@@ -502,6 +496,8 @@ def from_TPFs(
 
         if time_mask is not None:
             time_mask = np.copy(time_mask)[qual_mask]
+        # use or not the focus mask
+        focus_mask = focus_mask if do_focus_mask else None
```
christinahedges: What will happen when `focus_mask` is `None`?

src/psfmachine/tpf.py
christinahedges: This keyword also needs to be in the `__init__` function, right now it's just in the `from_TPFs` static method.

src/psfmachine/tpf.py
jorgemarpa: it passes the `None` value to the class `__init__` where is combined with the `time_mask`. If both none, then nothing is masked (see comment for lines 68 to 74: https://github.com/SSDataLab/psfmachine/pull/19#discussion_r652956679)

src/psfmachine/tpf.py
jorgemarpa: So I thought that `from_TPFs` will create the mask, but the actual masking happens in the `__init__` depending if `time_mask` and `focus_map` are valid or `None`

src/psfmachine/tpf.py
```diff
@@ -65,22 +65,15 @@ def __init__(
...
-            ),
-        )
-        if time_mask is None:
+        # combine focus mask and time mask if they exist
+        if time_mask is None and focus_mask is None:
+            self.time_mask = np.ones(len(self.time), bool)
+        elif time_mask is None and focus_mask is not None:
             self.time_mask = focus_mask
+        elif time_mask is not None and focus_mask is None:
+            self.time_mask = time_mask
```
jorgemarpa, line 74: Here is where the `time_mask` and `focus_mask` are applied depending on the case.

src/psfmachine/tpf.py
christinahedges: nice catch!

PR #22, opened 2021-06-22
FFIMachine

jorgemarpa:
A new subclass of `machine` to work with FFI data.

DONE:

- [x] Kepler FFIs 
- [x] save shape model (they can be used in `TPFMachine`)
- [x] save photometry
- [x] fancy plots

TODO:
- [x] fix the centroids being off for K2 data
- [ ] add ability to have different centroid offset per frame
- [x] unit testing
- [x] residuals
- [x] use multiple cadences with the same sparse arrays

Note: this is a sub-branch of `sparse-arrays`.


src/psfmachine/ffi.py
```diff
@@ -0,0 +1,924 @@
...
+__all__ = ["FFIMachine"]
+
+
+class FFIMachine(Machine):
+    """
+    Subclass of Machine for working with FFI data. It is a subclass of Machine
+    """
+
+    def __init__(
+        self,
```
christinahedges, line 30: Why isn't there are "flux" argument in the init? I don't understand how we can initialize an FFI machine without any flux.

src/psfmachine/ffi.py
```diff
@@ -0,0 +1,924 @@
...
+        channel=1,
+        quarter=5,
```
christinahedges: Kepler uses 
* "Quarters", "Modules", "Channels"
K2 uses 
* "Campaigns", "Modules", "Channels"
TESS uses
* "Sectors", "Cameras", "CCDs"
It would be nice if these keywords were a little flexible. For K2, there is no "quarter".

src/psfmachine/ffi.py
```diff
@@ -0,0 +1,924 @@
...
+            Quarter/Campagn nunmber to be used (for Kepler data).
+        """
+        self.column = kwargs["column"].ravel()
+        self.row = kwargs["row"].ravel()
+        self.ra = kwargs["ra"].ravel()
+        self.dec = kwargs["dec"].ravel()
+        # keep 2d image for easy plotting
+        self.flux_2d = kwargs["flux"]
+        # reshape flux and flux_err as [ntimes, npix]
+        self.flux = kwargs["flux"].reshape(kwargs["flux"].shape[0], -1)
```
christinahedges: It seems like you're putting the flux into kwargs. This is harder for the user, because 1) there's no prompt for flux, and if I don't pass flux this will break. 2) there is no docstring for flux. Let's be explicit. Make sure all the keyword arguments that are required in the init are required arguments for `__init__`

src/psfmachine/ffi.py
```diff
@@ -0,0 +1,924 @@
...
+            self.column,
+            self.row,
+            n_r_knots=n_r_knots,
+            n_phi_knots=n_phi_knots,
+            cut_r=cut_r,
+            rmin=rmin,
+            rmax=rmax,
+        )
+        self.meta = kwargs["metadata"]
+        self.channel = channel
```
christinahedges: This is where we could be more flexible and have some if statements for if we've passed any of 
* quarter
* campaign
* sector
* channel
* camera
* ccd

src/psfmachine/ffi.py
```diff
@@ -0,0 +1,924 @@
...
+            Channel number to be used
+        quarter : int
+            Quarter/Campagn nunmber to be used (for Kepler data).
+        """
+        self.column = kwargs["column"].ravel()
+        self.row = kwargs["row"].ravel()
+        self.ra = kwargs["ra"].ravel()
+        self.dec = kwargs["dec"].ravel()
+        # keep 2d image for easy plotting
+        self.flux_2d = kwargs["flux"]
```
christinahedges: why are there two flux attributes? This isn't seen anywhere else in `PSFmachine`?

src/psfmachine/ffi.py
christinahedges: You're spending memory here to have a second `flux` attribute. Why not have flux be 2D and then reshape it whenever necessary? or have flux be 1D, and reshape when necessary?

src/psfmachine/ffi.py
```diff
@@ -0,0 +1,924 @@
...
+        # reshape flux and flux_err as [ntimes, npix]
+        self.flux = kwargs["flux"].reshape(kwargs["flux"].shape[0], -1)
+        self.flux_err = kwargs["flux_err"].reshape(kwargs["flux_err"].shape[0], -1)
+        self.sources = kwargs["sources"]
+
```
christinahedges, line 104: Let's make sure to save the shape of the input flux somewhere (e.g. (1024x1100))

src/psfmachine/ffi.py
```diff
@@ -0,0 +1,924 @@
...
+            time=time.jd,
+            flux=flux,
+            flux_err=flux_err,
+            ra=ra,
+            dec=dec,
+            sources=sources,
+            column=column,
+            row=row,
+            channel=channel,
+            quarter=metadata["QUARTER"],
```
christinahedges: Won't exist for K2

src/psfmachine/ffi.py
```diff
@@ -0,0 +1,924 @@
...
+        self.channel = channel
+        self.quarter = quarter
+        self.wcs = wcs
+        self.flux_2d = kwargs["flux"]
+
+    def __repr__(self):
+        return f"FFIMachine (N sources, N times, N pixels): {self.shape}"
+
+    @staticmethod
+    def from_file(fname, channel=1, cutout_size=None, cutout_origin=[0, 0], **kwargs):
```
christinahedges: When I try to run this for a random TESS FFI it won't initialize (expected) but it breaks in a very weird place (no column attribute?) 

image

christinahedges: Same thing with a Kepler FFI 

image

christinahedges: Why do I need to specify the channel in `from_file`? It should be automatically populated by the file.

christinahedges (changes requested):
Hi @jorge, I'm going to stop my review for now because there seems to be a problem where it's not possible to initialize an FFI machine object from file. Take a look and let me know what you think, I can resume the review when we find the solution!

tests/test_ffimachine.py
```diff
@@ -0,0 +1,88 @@
...
+# import numpy as np
+import pytest
+from astropy.io import fits
+from astropy.table import Table
+from astropy.utils.data import get_pkg_data_filename
+
+from psfmachine import Machine, FFIMachine
+
+ffi_path = get_pkg_data_filename("./data/ffi_ch01_test.fits")
+ffi = FFIMachine.from_file(ffi_path, channel=1)
```
christinahedges: We need to move this inside the function. If this fails, if it's not in a test function we won't understand why it failed.

src/psfmachine/ffi.py
christinahedges: Could add static method for this e.g.
```python
@staticmethod
def flux_unwrap(self):
     return self.flux[apply_mask]
```
depends on what you want

tests/test_ffimachine.py
christinahedges:
```python
ffi = FFIMachine.from_file(ffi_path, extensions=1)
```

src/psfmachine/machine.py
```diff
@@ -46,6 +46,7 @@ def __init__(
         rmin=1,
         rmax=16,
         cut_r=6,
+        sparse_dist_lim=40,
```
christinahedges, line 49: Make sure to add documentation for this new keyword

src/psfmachine/machine.py
```diff
@@ -150,6 +151,7 @@ def __init__(
         self.rmin = rmin
         self.rmax = rmax
         self.cut_r = cut_r
+        self.sparse_dist_lim = sparse_dist_lim / 3600
```
christinahedges: use astropy units here?

src/psfmachine/machine.py
```diff
@@ -334,6 +336,7 @@ def _get_source_mask(
         lower_radius_limit=4.5,
         upper_flux_limit=2e5,
         lower_flux_limit=100,
+        correct_centroid_offset=True,
```
christinahedges, line 345: Need docstring for this

src/psfmachine/machine.py
```diff
@@ -502,30 +505,33 @@ def _get_source_mask(
...
-        if self.nsources * self.npixels < 1e7:
-            self._create_delta_arrays(centroid_offset=[ra_cent, dec_cent])
-        else:
-            self._create_delta_sparse_arrays(centroid_offset=[ra_cent, dec_cent])
+        # calculate image centroids and correct dra,ddec for offset.
+        if correct_centroid_offset:
+            self._get_centroids()
+            # print(self.ra_centroid_avg.to("arcsec"), self.dec_centroid_avg.to("arcsec"))
+            # re-estimate dra, ddec with centroid shifts, check if sparse case applies.
+            if self.nsources * self.npixels < 1e7:
```
christinahedges, line 525: Can you add a comment that this is hard coded based on your investigation? Maybe even link to the issue number so we can see the plots you made!

christinahedges (changes requested):
Hey @jorgemarpa, this is fantastic work. I requested some tiny changes (should take you 5 mins to just add a docstrings (they may just not have rendered on github?) otherwise this looks fantastic great work!

PR #25, opened 2021-07-23
[WIP] Aperture photometry with PRF model shape

jorgemarpa:
This PR implements aperture photometry using the PRF model build in `psfmachine`. It includes:

- Aperture photometry for user-defined and optimal size.
- Flux metrics: flux completeness and crowding
- Diagnostic plots for metrics
- Enhance `fit_lightcurve` to include aperture photometry
- Enhanced `plot_tpf` to include SAP light curve and aperture mask


jorgemarpa:
This is ready to be reviewed @christinahedges .

Note: I moved the code that creates the metadata dictionary for each light curve into a dedicated function to simplify `fit_lightcurves()`. This also helps to pass flake8.

christinahedges:
Hi @jorgemarpa, here are some top level comments before I review the code here:

1. It seems like there are some functions in `machine.py` that might be better in `utils.py` or perhaps a bespoke aperture module. e.g. the `compute_FLFRSAP` function isn't relevant to the `Machine` class, so let's move as much as possible to a different module. This is so that each module is easy to read and small. It will also help us with the documentation.
2. There needs to be some update to the docs for this PR, in `docs/` there are several mark down files, make sure this new functionality is added to the mkdocs file so that it appears in the API docs! When moving some of the functions to a new module, you may need to make a new `.md` file in `docs/`
3. This is a point for discussion, do the aperture photometry tools belong in `Machine`, or do they belong in `TPFMachine`? It may be that you are exactly right and these belong in `Machine`, but it is worth just thinking about carefully before we commit to it. 
4. In the doc string for `fit_lightcurves` (and wherever else appropriate) can we have some explicit text about what is being created, e.g. 
`# fit_lightcurves creates the lc attribute, which contains X, PSF photometry is available in the Y attribute, SAP in the Z attribute. They are in format A`
or similar, so people know how to get into the relevant light curves. 

Take a look at these top level comments and then I'll review the PR! On a cursory look it looks great!

jorgemarpa:
Thanks for the feedback!

> Hi @jorgemarpa, here are some top level comments before I review the code here:
> 
> 1. It seems like there are some functions in `machine.py` that might be better in `utils.py` or perhaps a bespoke aperture module. e.g. the `compute_FLFRSAP` function isn't relevant to the `Machine` class, so let's move as much as possible to a different module. This is so that each module is easy to read and small. It will also help us with the documentation.

This make sense.

> 2. There needs to be some update to the docs for this PR, in `docs/` there are several mark down files, make sure this new functionality is added to the mkdocs file so that it appears in the API docs! When moving some of the functions to a new module, you may need to make a new `.md` file in `docs/`

Ok, I'll update those files.

> 3. This is a point for discussion, do the aperture photometry tools belong in `Machine`, or do they belong in `TPFMachine`? It may be that you are exactly right and these belong in `Machine`, but it is worth just thinking about carefully before we commit to it.

IMO this function belongs to `machine`, so it is independent of how the data is loaded. In this way, the SAP can be computed for data loaded from TPFs, FFIs, or any other module we added to the package. 

> 4. In the doc string for `fit_lightcurves` (and wherever else appropriate) can we have some explicit text about what is being created, e.g.
>    `# fit_lightcurves creates the lc attribute, which contains X, PSF photometry is available in the Y attribute, SAP in the Z attribute. They are in format A`
>    or similar, so people know how to get into the relevant light curves.

Sounds good.

> 
> Take a look at these top level comments and then I'll review the PR! On a cursory look it looks great!

jorgemarpa:
@christinahedges merge conflicts solved! 
Ready for review

src/psfmachine/machine.py
```diff
@@ -1247,47 +1295,278 @@ def fit_model(self, fit_va=False):
 
         return
 
+    # aperture photometry functions
+    def _create_aperture_mask(self, percentile=50):
```
christinahedges: We might want to have some information about where this is from. This is from Kepler apertures right?

src/psfmachine/machine.py
```diff
@@ -1247,47 +1295,278 @@ def fit_model(self, fit_va=False):
...
+        ):
+            self.sap_flux[:, sdx] = self.flux[:, self.aperture_mask[sdx]].sum(axis=1)
+            self.sap_flux_err[:, sdx] = (
+                np.power(self.flux_err[:, self.aperture_mask[sdx]], 2).sum(axis=1)
+                ** 0.5
+            )
+
+        return
+
+    def _optimize_aperture(
```
christinahedges: This is making machine a little large, should we put these in an `apertures` module?

src/psfmachine/machine.py
```diff
@@ -1247,47 +1295,278 @@ def fit_model(self, fit_va=False):
...
+        # compute flux metrics. Have to round to 10th decimal due to floating point
+        self.FLFRCSAP = np.round(
+            compute_FLFRCSAP(self.mean_model, self.aperture_mask), 10
+        )
+        self.CROWDSAP = np.round(
+            compute_CROWDSAP(self.mean_model, self.aperture_mask), 10
+        )
+
+    def compute_aperture_photometry(
+        self, aperture_size="optimal", target_complet=0.9, target_crowd=0.9
```
christinahedges: Is this a typo? Should it be `target_complete`?

src/psfmachine/machine.py
```diff
@@ -1247,47 +1295,278 @@ def fit_model(self, fit_va=False):
...
+            self._optimize_aperture(
+                target_complet=target_complet, target_crowd=target_crowd
+            )
+        else:
+            self._create_aperture_mask(percentile=aperture_size)
+
+        self.sap_flux = np.zeros((self.flux.shape[0], self.nsources))
+        self.sap_flux_err = np.zeros((self.flux.shape[0], self.nsources))
+
+        for sdx in tqdm(
```
christinahedges, line 1374: I think you could improve speed here slightly, depending on how you're storing the apertures, by not looping over each source.
Something like:
`(self.flux * self.aperture_mask.ravel()).sum(axis=1)`

But this is a minor comment.

src/psfmachine/machine.py
```diff
@@ -1247,47 +1295,278 @@ def fit_model(self, fit_va=False):
...
+        ):
+            optim_params = {
+                "percentile_bounds": percentile_bounds,
+                "target_complet": target_complet,
+                "target_crowd": target_crowd,
+                "max_iter": max_iter,
+                "psf_models": self.mean_model,
+                "sdx": sdx,
+            }
+            minimize_result = optimize.minimize_scalar(
```
christinahedges: `optimize` as a name might not be very descriptive, but having this in an apertures module might make it a little more descriptive.

src/psfmachine/tpf.py
```diff
@@ -441,6 +529,132 @@ def save_shape_model(self, output=None):
 
         return
 
+    def _aperture_mask_to_2d(self):
+        """
+        Convert 1D aperture mask into 2D to match the shape of TPFs. This 2D aperture
```
christinahedges: I don't understand this function, aren't aperture masks always 2D? (they are images)

src/psfmachine/utils.py
```diff
@@ -343,3 +343,105 @@ def solve_linear_model(
...
+        mask = arr.data < limit
+    # reconstruct sparse array
+    masked_arr = sparse.csr_matrix(
+        (arr.data[mask], (nonz_idx[0][mask], nonz_idx[1][mask])),
+        shape=arr.shape,
+    ).astype(bool)
+    return masked_arr
+
+
+def compute_FLFRCSAP(psf_models, aperture_mask):
```
christinahedges: Probably in aperture module?

src/psfmachine/utils.py
```diff
@@ -343,3 +343,105 @@ def solve_linear_model(
...
+    FLFRCSAP: numpy.ndarray
+        Completeness metric
+    """
+    return np.array(
+        psf_models.multiply(aperture_mask.astype(float)).sum(axis=1)
+        / psf_models.sum(axis=1)
+    ).ravel()
+
+
+def compute_CROWDSAP(psf_models, aperture_mask, idx=None):
```
christinahedges: Probably also would be good in aperture module

christinahedges (commented):
Hi @jorgemarpa this seems great! I left a few comments, my biggest one is it would be nice to have an `apertures` module to take some of this new material out of `utils` and `Machine`, as there's a lot of new content. You can also point back to your kepler apertures tool to show where it came from.

christinahedges:
Oh also last comment if we're merging this we should update the version number!

src/psfmachine/machine.py
jorgemarpa: This won't work because `aperture_mask.shape = [n_sources, n_pixels]` and `flux.shape = [n_times, n_pixels]`.

src/psfmachine/tpf.py
jorgemarpa: I update the docstring to make it more explicative

jorgemarpa:
Ok, I made the updates accordingly. 
* I moved all aperture-related code to `aperture_utils.py`, with the exception of `compute_aperture_photometry` that remains in `machine.py` as the interface and main method to compute aperture photometry.
* I also update the docstrings and `md` files accordingly.
* Created a `CHANGES.rst` to keep track of version updates and bumped up the version to `1.1.0`.

src/psfmachine/aperture_utils.py
```diff
@@ -0,0 +1,378 @@
...
+    ratio = psf_models.multiply(1 / psf_models.sum(axis=0)).tocsr()
+    if idx is None:
+        return np.array(
+            ratio.multiply(aperture_mask.astype(float)).sum(axis=1)
+        ).ravel() / aperture_mask.sum(axis=1)
+    else:
+        return ratio[idx].toarray()[0][aperture_mask].sum() / aperture_mask.sum()
+
+
+def aperture_mask_to_2d(machine):
```
christinahedges: Maybe belongs in TPFMachine like this? Or change the function signature to take in the TPFs and the meta dictionary?

Maybe consider:

```Python
def plotting_func(self):
    dictionary = self._aperture_mask_to_2d()
    ... 
```

or

```Python
def plotting_func(self):
    dictionary = aperture_utils._aperture_mask_to_2d(self.tpfs, self.tpf_meta)
    ... 
```

src/psfmachine/aperture_utils.py
```diff
@@ -0,0 +1,378 @@
...
+
+import numpy as np
+import matplotlib.pyplot as plt
+import astropy.units as u
+from scipy import optimize
+from tqdm import tqdm
+
+
+# aperture photometry functions
+def create_aperture_mask(machine, percentile=50):
```
christinahedges: Maybe this does belong in machine?

src/psfmachine/aperture_utils.py
```diff
@@ -0,0 +1,378 @@
...
+    ax.plot(cut, crowd, label=r"CROWDSAP", c="tab:green")
+    if optimal_percentile:
+        ax.axvline(optimal_percentile, c="tab:red", label="optimal")
+    ax.set_xlabel("Percentile")
+    ax.set_ylabel("Metric")
+    ax.legend()
+    return ax
+
+
+def estimate_source_centroids_aperture(machine):
```
christinahedges: Probably also belongs in `Machine` after all

jorgemarpa:
I rearrange some of the aperture functions to make all `aperture.py` machine agnostic. 

Note: `flake8` is complaining that `TPFMachine.plot_tpf()` is too complex (15 score, and max allowed is 14).

src/psfmachine/tpf.py
```diff
@@ -245,6 +308,16 @@ def plot_tpf(self, tdx):
         r += tpf.row
```
christinahedges, line 308:
```suggestion
        r, c = np.mgrid[: tpf.shape[1], : tpf.shape[2]] + np.htsack([tpf.row, tpf.column])
```

src/psfmachine/aperture_utils.py
jorgemarpa: It's now `machine` agnostic

jorgemarpa:
I change flakes8's complexity max level to 15.

christinahedges:
Jorge and I agree that increasing complexity limit is reasonable here! LGTM I'm merging.

PR #31, opened 2021-08-10
[ISSUE #30]: computing radec from WCS of each TPF

### LINKED ISSUE #30: pixel to WC doesn't match Gaia source coordinate for some TPFs
Some Gaia source coordinates do not align with TPFs data. See the following figure:

image

The current `tpf._wcs_from_tpfs()` uses the WCS from the first TPF of the stack instead of each solution.
https://github.com/SSDataLab/psfmachine/blob/89190e150bdc697fafd363c1a1e41fa37c4345d2/src/psfmachine/tpf.py#L843-L850

jorgemarpa:
This PR fixes Issue #30. 
The pixel to world coordinate conversion is done by using the WCS solution of each TPF.

Figure to compare with the one posted in #30.
image

closes #30

### DIFF
--- a/src/psfmachine/tpf.py
+++ b/src/psfmachine/tpf.py
@@ -828,26 +828,23 @@ def _wcs_from_tpfs(tpfs):
     dec: numpy.ndarray
         Array with declination values per pixel
     """
-    # calculate x,y grid of each pixel
-    locs = np.hstack(
-        [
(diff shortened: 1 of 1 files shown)


christinahedges:
This is great work! Thanks @jorgemarpa for finding and fixing this!

PR #32, opened 2021-09-14
use shape-model files from zenodo

jorgemarpa:
I created a Zenodo repo with the shape model files: https://zenodo.org/record/5504503#.YT_pz9MzY8M

As a tar.gz the file is 1.8 MB, it could be included in the GitHub repo. BUT I think Zenodo makes more sense for the version controlling and DOI features. 

This PR include improvements in `load_shape_model()` method, when `input=None` then it download from Zenodo and uncompress the tar.gz file into `src/psfmachine/data/fff/`. This happens only when the FITS file and/or the tar.gz aren't on disk.


jorgemarpa:
@christinahedges  this is ready to be reviewed

src/psfmachine/tpf.py
```diff
@@ -296,18 +299,40 @@ def load_shape_model(self, input=None, plot=False):
...
         # By default we will load PRF model from FFI when this are ok.
-        # for now this function only works when a file is provided
         if input is None:
-            raise NotImplementedError(
-                "Loading default model not implemented. Please provide input file."
+            # raise NotImplementedError(
+            #     "Loading default model not implemented. Please provide input file."
+            # )
+            input = "%s/data/ffi/ch%02i/%s_ffi_shape_model_ch%02i_q%02i.fits" % (
+                PACKAGEDIR,
```
christinahedges: I'm going to merge this as is but I do find f strings to be a little easier to read!

src/psfmachine/tpf.py
```diff
@@ -296,18 +299,40 @@ def load_shape_model(self, input=None, plot=False):
...
+                self.tpf_meta["channel"][0],
+                self.tpf_meta["mission"][0],
+                self.tpf_meta["channel"][0],
+                self.tpf_meta["quarter"][0],
             )
+            if not os.path.isfile(input):
+                # if dile doesnt exist, download file bundle from zenodo:
+                tar_file = "%s/data/Kepler_FFI_PRFmodels_v1.0.tar.gz" % PACKAGEDIR
```
christinahedges: e.g. this is an f string
```suggestion
                tar_file = f"{PACKAGEDIR}/data/Kepler_FFI_PRFmodels_v1.0.tar.gz"
```

src/psfmachine/tpf.py
```diff
@@ -296,18 +299,40 @@ def load_shape_model(self, input=None, plot=False):
...
+                if not os.path.isfile(tar_file):
+                    url = "https://zenodo.org/record/5504503/files/Kepler_FFI_PRFmodels_v1.0.tar.gz?download=1"
+                    print("Downloading bundle files from: ", url)
```
christinahedges: Can we have this as a logging message so it is easy to silence?
```suggestion
log.info(f"Downloading bundle files from: {url}")
```

src/psfmachine/tpf.py
```diff
@@ -296,18 +299,40 @@ def load_shape_model(self, input=None, plot=False):
...
+                        tar_file, "wb"
+                    ) as out_file:
+                        out_file.write(response.read())
+                # unpack
+                with tarfile.open(tar_file) as f:
+                    f.extractall("%s/data/ffi/" % (PACKAGEDIR))
         # check if file exists and is the right format
         if not os.path.isfile(input):
             raise FileNotFoundError("No shape file: %s" % input)
+        print("Using shape model from: ", input)
```
christinahedges: Same as above, logging messages are easy to silence and print statements are less easy!

src/psfmachine/tpf.py
```diff
@@ -296,18 +299,40 @@ def load_shape_model(self, input=None, plot=False):
...
+                self.tpf_meta["channel"][0],
+                self.tpf_meta["mission"][0],
+                self.tpf_meta["channel"][0],
+                self.tpf_meta["quarter"][0],
             )
+            if not os.path.isfile(input):
+                # if dile doesnt exist, download file bundle from zenodo:
+                tar_file = "%s/data/Kepler_FFI_PRFmodels_v1.0.tar.gz" % PACKAGEDIR
+                if not os.path.isfile(tar_file):
+                    url = "https://zenodo.org/record/5504503/files/Kepler_FFI_PRFmodels_v1.0.tar.gz?download=1"
```
christinahedges: Are you going to have the versions updated on zenodo sometimes? There's a version number in the filename, so just thought it would be worth checking in.

christinahedges (changes requested):
This looks great! If you can change those print statements to logging messages I will merge! There should already be logging in place in PSFmachine.

src/psfmachine/tpf.py
jorgemarpa: I probably won't. I'll actually let Zenodo do the version controlling here, then the `v1.0` in the name doesn’t make much more sense. I'll keep it to be consistent with the file name in Zenodo.

jorgemarpa:
Ok, I made the f-string changes, although for some ling lines have to find some decent line breakers.  
Made self-consistent the local tarball file name with the URL address.

About the logger. We currently don't have a logger, so I added it. I took the chance to fix this also for `ffi.py` that also had a couple of `prints` that were replaced by `log.info`.

I also fixed a directory not-found bug that happened when downloading the Zenodo file in a clean repo.

PS: one day I'll have to update all the %-formatting to f-string 😄

christinahedges:
Hi @jorgemarpa

This looks really great. The only thing I think is missing is a unit test that checks it's possible to download these files, and load them into a TPF machine. I think this would be a pretty simple test, and you can mark it with the `@pytest.mark.remote_data` decorator to show that the test requires internet connection. 

I think if we don't add it in here, we won't test it, and it'll be much harder to debug down the line if this fails!

jorgemarpa:
Done!

PR #35, opened 2021-11-09
Using `pos_corr`s as time correctors

jorgemarpa:
This PR adds the option to use `pos_corr`s vectors to fit the time model. With this, the time model can be built either using a time polynomial or the `pos_corr`s. This is controlled by the attribute `self.time_corrector` that can be `"pos_corr"` or `"polynomial"` (default). The `pos_corr` vectors are smoothened with a Gaussian kernel to reduce noise. 

Example of `pos_corr` vectors for quarter 2 (one of the worst behave)
image
image
Black line is the median value of the vectors.
Red line is the smooth version
Red dots are the binned version.

It also adds the option to modify the spacing between knots in the Cartessian design matrix which can be "linear" or "sqrt" (default), controlled with the attribute `self.cartesian_knot_spacing`.

Here two examples showing `pos_corr` vs `polynomial`
image
image
The light curves that uses `pos_corr` (red) have significantly less long-term trends due to the instrument compared with the polynomial version (in blue).

And in general (statistics over ~4000 light curves) using `pos_corr` keeps a low level of CDPP and "flatten" light curves (e.g see Linear Trend distribution in the following figure)
image


src/psfmachine/machine.py
```diff
@@ -681,12 +721,62 @@ def _time_bin(self, npoints=200):
...
+
+            # do poscorr binning
+            pc1_bin = np.vstack(
+                [np.median(t1, axis=0) for t1 in np.array_split(pc1_smooth, breaks)]
+            ).ravel()[:, None] * np.ones(fm.shape)
+            pc2_bin = np.vstack(
+                [np.median(t1, axis=0) for t1 in np.array_split(pc2_smooth, breaks)]
+            ).ravel()[:, None] * np.ones(fm.shape)
+
+            return (
```
christinahedges, line 768: I can't find a docstring for all these returned variables

src/psfmachine/machine.py
```diff
@@ -617,21 +633,45 @@ def _time_bin(self, npoints=200):
...
+        # if using poscorr, find and add discontinuity in poscorr data
+        if hasattr(self, "pos_corr1") and self.time_corrector == "pos_corr":
+            # take the scene-median poscorr
+            mpc1 = np.nanmedian(self.pos_corr1, axis=0)
+            mpc2 = np.nanmedian(self.pos_corr2, axis=0)
+
+            # find poscorr discontinuity in each axis
+            grads1 = np.gradient(mpc1, self.time)
+            grads2 = np.gradient(mpc2, self.time)
+            splits1 = np.where(grads1 > 7 * grads1.std())[0]
```
christinahedges, line 661: Put a comment that we're hard coding this somewhere

src/psfmachine/machine.py
```diff
@@ -617,21 +633,45 @@ def _time_bin(self, npoints=200):
...
+            splits2 = np.where(grads2 > 7 * grads2.std())[0]
+            # merging breaks
+            splits = np.unique(np.concatenate([splits, splits1[1::2], splits2[1::2]]))
+            del grads1, grads2, splits1, splits2
+
```
christinahedges, line 666: I don't understand what `splits_a` and `_b` is, and I don't know why there is a +100 here

src/psfmachine/machine.py
```diff
@@ -617,21 +633,45 @@ def _time_bin(self, npoints=200):
...
         splits_a = splits[:-1] + 100
         splits_b = splits[1:]
         dsplits = (splits_b - splits_a) // npoints
-        breaks = []
+        # breake has the bin centers
+        # first break is at cadance 50
+        breaks = [50]
```
christinahedges: This seems fairly justified, but is hard coded, maybe you want it as a percentage? (e.g. think about how short Q0 is)

src/psfmachine/machine.py
```diff
@@ -617,21 +633,45 @@ def _time_bin(self, npoints=200):
...
         for spdx in range(len(splits_a)):
             breaks.append(splits_a[spdx] + np.arange(0, dsplits[spdx] - 1) * npoints)
+        # we include 50-to-last cadecence
+        breaks.append(splits[-1] - 50)
```
christinahedges: This will break if `splits` has any values less than 50 or more than len(time) - 50.

src/psfmachine/machine.py
```diff
@@ -681,12 +721,62 @@ def _time_bin(self, npoints=200):
 
         tm = ((tm - tm.mean()) / (tm.max() - tm.mean()))[:, None] * np.ones(fm.shape)
 
+        # poscor
+        if hasattr(self, "pos_corr1") and self.time_corrector == "pos_corr":
+            # we smooth the poscorr with a Gaussian kernel and 12 cadence window
+            # (6hr-CDPP) to not introduce too much noise, the smoothing is aware
```
christinahedges, line 732: Brilliant, great comment.

src/psfmachine/machine.py
```diff
@@ -617,21 +633,45 @@ def _time_bin(self, npoints=200):
...
-        breaks = []
+        # breake has the bin centers
+        # first break is at cadance 50
+        breaks = [50]
         for spdx in range(len(splits_a)):
             breaks.append(splits_a[spdx] + np.arange(0, dsplits[spdx] - 1) * npoints)
+        # we include 50-to-last cadecence
+        breaks.append(splits[-1] - 50)
         breaks = np.hstack(breaks)
+        # breaks = np.unique(np.hstack([breaks, splits[1:-1]]))
```
christinahedges: I think you can maybe the previous few lines and check there's not a simpler implementation

src/psfmachine/machine.py
```diff
@@ -681,12 +721,62 @@ def _time_bin(self, npoints=200):
...
+            # we smooth the poscorr with a Gaussian kernel and 12 cadence window
+            # (6hr-CDPP) to not introduce too much noise, the smoothing is aware
+            # of focus-change breaks
+            pc1_smooth = []
+            pc2_smooth = []
+            for i in range(1, len(splits)):
+                pc1_smooth.extend(
+                    gaussian_filter1d(
+                        mpc1[splits[i - 1] : splits[i]],
+                        self.poscorr_filter_size,
```
christinahedges, line 746: Can we add a test that this still runs if `self.poscorr_filter_size=0`

src/psfmachine/machine.py
```diff
@@ -713,19 +818,36 @@ def build_time_model(self, plot=False):
...
+            # Cartesian spline with time dependence
+            A3 = sparse.hstack(
+                [
+                    A2,
+                    A2.multiply(time_binned.ravel()[:, None]),
+                    A2.multiply(time_binned.ravel()[:, None] ** 2),
+                    A2.multiply(time_binned.ravel()[:, None] ** 3),
+                ],
+                format="csr",
+            )
```
christinahedges: It seems like we need some helper functions to abstract out the matrix building

christinahedges (commented):
I think this is looking pretty good. My biggest comment is that we should refactor some of the time correction stuff in machine into smaller helper functions, for readability. This will help when there are multiple types of time correction to use.

src/psfmachine/machine.py
jorgemarpa: this makes sense. I'll add a helper function in `utils.py` to take care of this

src/psfmachine/machine.py
jorgemarpa: these are in lines 638-642

src/psfmachine/machine.py
jorgemarpa: fixed, same as above, I change it to a percentage of the total number of cadences

src/psfmachine/machine.py
jorgemarpa: this is a way to ensure no knots fall in the poscorr/time discontinuities.
I added comments to those lines explaining.
I also try a simple `np.arange(0.01 * nt, 0.99 * nt, npoints)`, but this way knots can fall right on top of a cadence break and lead to funky corrected light curves.

src/psfmachine/machine.py
jorgemarpa: see answer above for `splits_a` and `splits_b`

PR #36, opened 2021-11-10
K2 SuperStamps class

jorgemarpa:
The new module is in `superstamp.py`. This new module of `psfmachine` allows us to work with K2 SuperStamps data (one FITS file per cadence as created by Cody et al. 2018).

It includes 2 main new functionalities:
* build shape models per frame
* fit the shape model on each frame and get light curves

It also includes some plotting tools:
* plots the super stamp at a given frame
* creates a movie of the shape models


src/psfmachine/superstamp.py
```diff
@@ -0,0 +1,533 @@
...
+        for tdx in tqdm(range(self.nt), desc="Building shape model per frame"):
+            fig = self.build_shape_model(frame_index=tdx, plot=plot, **kwargs)
+            self.mean_model_frame.append(self.mean_model)
+            if plot:
+                fig.canvas.draw()  # draw the canvas, cache the renderer
+                image = np.frombuffer(fig.canvas.tostring_rgb(), dtype="uint8")
+                image = image.reshape(fig.canvas.get_width_height()[::-1] + (3,))
+                images.append(image)
+                plt.close()
+            self.source_mask = org_sm
```
christinahedges, line 118: Can you add a comment here that we have to reset the source mask because the sources may move slightly (e..g K2)

src/psfmachine/superstamp.py
```diff
@@ -0,0 +1,533 @@
...
+            self.werrs_frame[tdx, nodata] *= np.nan
+
+    @staticmethod
+    def from_file(fname, magnitude_limit=18, dr=2, **kwargs):
+        """
+        Reads data from files and initiates a new FFIMachine class.
+        Parameters
+        ----------
+        fname : string or list of strings
+            Name of the FITS files to be parsed.
```
christinahedges: Is this. a specific file? I think it might be, so worth adding in the comments that only certain files will work?

src/psfmachine/superstamp.py
```diff
@@ -0,0 +1,533 @@
...
+            nodata = np.asarray(self.source_mask.sum(axis=1))[:, 0] == 0
+            # These sources are poorly estimated
+            nodata |= (self.mean_model_frame[tdx].max(axis=1) > 1).toarray()[:, 0]
+            self.ws_frame[tdx, nodata] *= np.nan
+            self.werrs_frame[tdx, nodata] *= np.nan
+
+    @staticmethod
+    def from_file(fname, magnitude_limit=18, dr=2, **kwargs):
+        """
+        Reads data from files and initiates a new FFIMachine class.
```
christinahedges: I think you mean SSMachine here?

christinahedges: Can you add a comment about how to use this? Is it one filename per frame? How do I load all the superstamp files?

src/psfmachine/superstamp.py
```diff
@@ -0,0 +1,533 @@
...
+            dec.ravel(),
+            sources,
+            column.ravel(),
+            row.ravel(),
+            wcs=wcs,
+            meta=metadata,
+            **kwargs,
+        )
+
+    def fit_lightcurves(
```
christinahedges, line 267: Is this not an inherited method from the subclass?

src/psfmachine/superstamp.py
```diff
@@ -0,0 +1,533 @@
...
+            fig.canvas.draw_idle()
+
+        interact(update, t=(0, self.flux_2d.shape[0] - 1, 1))
+
+        return
+
+
+def _load_file(fname):
+    """
+    Helper function to load FFI files and parse data.
```
christinahedges: I think these are specifically the superstamp files only, let's include a description of the expected file format (e.g. one image per file, n files, etc)

christinahedges (commented):
This is looking in really good shape. I've added a few comments. Biggest comment I have is it's not clear from the code how i should interact with it. How do I pass a list of multiple files (I think that's the superstamp format). Otherwise, this looks in great shape.

src/psfmachine/superstamp.py
jorgemarpa: I update the docstrings to explain this

src/psfmachine/superstamp.py
jorgemarpa: Not quite, this version has the option to do:
- mean-PSF photometry
- mean-PSF photometry with VA
- cadence-PSF photometry
- SAP
Also, `SSMAchine` is a subclass of `FFIMachine` and `Machine`, not `TPFMachine`.

jorgemarpa: I do want to move all the code that creates the LC metadata out of `tpf.py` and `superstamp.py` to a dedicated function (maybe in `utils.py`) to remove repeated code. But I want to do that in a separate PR along with more code cleaning I want to do. 
What do you think?

src/psfmachine/superstamp.py
jorgemarpa: better docstrings now

jorgemarpa:
Last updates include:

- Removing off-sensor sources for super stamps, e.g. M35 is not a "square" image. Also, allows for sources up to 2 pixels from TPF border. 
- Reads `pos_corrs` from super stamps headers and fixes a bug when parsing MJD from headers.
- When building a time model:
    - Use `self.time_corrector = "centroids"` as basis option
    - downsample instead of binning the `pos_corrs` and `time` to improve value representativity in the K2 case

src/psfmachine/superstamp.py
```diff
@@ -0,0 +1,572 @@
...
+        if sap:
+            self.build_shape_model(plot=plot)
+            self.compute_aperture_photometry(
+                aperture_size="optimal", target_complete=1, target_crowd=1
+            )
+
+        # do mean-PSF photometry and time model if asked
+        if fit_mean_shape_model:
+            self.build_shape_model(plot=plot)
+            self.build_time_model(plot=plot, donwsample=True)
```
christinahedges:
```suggestion
            self.build_time_model(plot=plot, downsample=True)
```

src/psfmachine/superstamp.py
```diff
@@ -0,0 +1,572 @@
...
+        # do SAP first
+        if sap:
+            self.build_shape_model(plot=plot)
+            self.compute_aperture_photometry(
+                aperture_size="optimal", target_complete=1, target_crowd=1
+            )
+
+        # do mean-PSF photometry and time model if asked
+        if fit_mean_shape_model:
+            self.build_shape_model(plot=plot)
```
christinahedges:
```suggestion
```

christinahedges: Duplicated

src/psfmachine/machine.py
```diff
@@ -679,55 +689,72 @@ def _time_bin(self, npoints=200):
...
-        fm_raw = np.asarray(
-            [
-                (
-                    sparse.csr_matrix(self.uncontaminated_source_mask)
-                    .multiply(self.flux[ms[tdx]].mean(axis=0))
-                    .data
-                )
-                for tdx in range(len(ms))
+        if not downsample:
+            tm = np.vstack(
```
christinahedges, line 695: These all have variable names that are hard to understand (`ta`, `tm`, `ms` etc), can we have some more comments in this part of the code so we know what's going on here?

src/psfmachine/superstamp.py
christinahedges: I think this sounds like a great idea!

christinahedges (commented):
I think this looks great. I believe that in a future PR we might try to consolidate SSMachine and FFIMachine, but I'm not going to comment on that here because I think that is in the future.

This seems really close to being ready to go. If we can have a few extra comments in that one section, I'll merge.

src/psfmachine/superstamp.py
jorgemarpa: fixed

src/psfmachine/machine.py
jorgemarpa: I added some comments that tell what those variables are

jorgemarpa:
I complete some docstrings and comments, plus I added the **cutout** option borrowed from `FFIMachine` (this convinced me more about merging `FFIMachine` with `SSMachine`) but we'll leave that for a different PR

PR #38, opened 2021-11-23
[#37] Fix SAP bug when mean_model is zero for a source

### LINKED ISSUE #37: SAP return ill-defined apertures for sources with `mean_model = 0`
When using `fit_lightcurves()` with `iter_negative=True` sources that get negative fluxes are clipped from the fitting by setting its `mean_model`  to zero. This problem happens especially when the PSF fitting struggles separating highly blended sources, 2 sources within a fraction of a pixel distance, and having similar brightness. In these cases, usually, the fitting returns one source with positive flux and the other with negative flux. Because of the negative flux, one of the light curves gets rejected. Here is an example:

Screen Shot 2021-11-23 at 10 43 30

Here, only the second source has PSF photometry, while the first has only SAP.
Additionally, the aperture mask for the SAP light curve of the first source is wrong, in this case, set to all pixels in the TPF. 

NOTE: this issue is only tracking the SAP bug.

jorgemarpa:
This PR fixes #37.
I moved the sap photometry in `fit_lightcurve()` before fitting the PSF photometry and iteration over negatives because the SAP photometry is independent of the PSF one. it only uses the shape model. This avoids the problem of using a modified `mean_model`.
I also make sure that there are no "all-pixel" apertures when a source has `mean_model = 0`. In this case (usually for sources outside the TPF), an aperture can't be defined and is set to 0.

### DIFF
--- a/src/psfmachine/machine.py
+++ b/src/psfmachine/machine.py
@@ -1327,6 +1327,8 @@ class Machine(object):
         )
         # create aperture mask
         self.aperture_mask = np.array(self.mean_model >= cut[::, None])
+        # make sure there are no all-pixel apertures due to `mean_model[i] = 0`
+        self.aperture_mask[self.aperture_mask.sum(axis=1) == self.npixels] = 0
         # compute flux metrics. Have to round to 10th decimal due to floating point
         self.FLFRCSAP = np.round(
             compute_FLFRCSAP(self.mean_model, self.aperture_mask), 10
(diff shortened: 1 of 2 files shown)


src/psfmachine/machine.py
```diff
@@ -1327,6 +1327,8 @@ def create_aperture_mask(self, percentile=50):
         )
         # create aperture mask
         self.aperture_mask = np.array(self.mean_model >= cut[::, None])
+        # make sure there are no all-pixel apertures due to `mean_model[i] = 0`
+        self.aperture_mask[self.aperture_mask.sum(axis=1) == self.npixels] = 0
```
christinahedges: Should probably be `False` not `0`

christinahedges:
```suggestion
        self.aperture_mask[self.aperture_mask.sum(axis=1) == self.npixels] = False
```

christinahedges (approved):

christinahedges:
LGTM! Merging!

PR #42, opened 2021-12-09
Gaia query

### LINKED ISSUE #41: Gaia query with multiple cone-serch points takes long time
https://github.com/SSDataLab/psfmachine/blob/390d89c7f259529e4cd08dc82c5b434d401f0fc4/src/psfmachine/utils.py#L13

When used by `TPFMachine.from_TPFs()` a single query contains a cone search with multiple points/circles. This week the Gaia server started having issues with this type of query. The wait times increased drastically, sometimes never finishing, for more than 50 points/circles.

Also, need to cache the results to disk instead of memory, to avoid repeating long queries after rebooting python kernel

jorgemarpa:
This PR fixes issue #41 by changing the order of point/circle in the ADQL Gaia query following the recommendation here (green box):
https://www.cosmos.esa.int/web/gaia-users/archive/combine-with-other-data#CrossMatchUserTableFlag

Also adds disk cache instead of memory. Results are cached for 30 days with a size limit of 1GB.

Closes #41

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -382,6 +382,14 @@ python-versions = ">=2.6, !=3.0.*"
 graph = ["objgraph (>=1.7.2)"]
 
 [[package]]
+name = "diskcache"
+version = "5.3.0"
+description = "Disk Cache -- Disk and file backed persistent cache."
+category = "main"
+optional = false
+python-versions = ">=3"
+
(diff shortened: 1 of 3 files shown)


christinahedges (approved):
LGTM!

PR #43, opened 2021-12-09
use self.time_radius when building cartesian DM

### LINKED ISSUE #40: plot_time_model() does not use self.time_radius
https://github.com/SSDataLab/psfmachine/blob/390d89c7f259529e4cd08dc82c5b434d401f0fc4/src/psfmachine/machine.py#L935

jorgemarpa:
This PR fixes #40 and makes sure that `self.time_radius` is always used when doing `_make_A_cartesian()`


christinahedges (approved):
LGTM!

PR #45, opened 2022-01-05
Fit background pixels

jorgemarpa:
This PR adds the functionality of modeling the background pixels to remove the "rolling band" effect seen in Kepler data.

New Flags/Attributes
- `fit_bkg` argument is included in `TPFMachine.from_TPFs()` to turn on/off background fitting. This is also an attribute of `TPFMachine` for internal use.
- `bkg_substracted` boolean that controls if data is or has been background subtracted. This variable is for internal use.

New methods:

- `build_background_model()`: uses [`kbackground`](https://github.com/SSDataLab/kbackground) to model the background.
- `plot_background_model()`: creates a diagnostic plot to see the data and model.

Updated methods:

- `fit_lightcurves()`: includes the step that models the background.
- `fit_model()`: includes the step that removes the background from data.


jorgemarpa:
`pytest` failed because `kbackground` uses `python >= 3.8` and `psfmachine` uses `python ^ 3.7` 
@christinahedges is there any particular reason to not include `python 3.7`?

christinahedges:
I'll update the version for kbackground and see what happens, but I have a nagging feeling there might actually have been a reason...

jorgemarpa:
I saw you updated the python version for `kbakcground`, but I'm still having problems adding it to `psmachine`.  It is still complaining about the python version and it does not make sense to me. Here is the error:

```
poetry add kbackground==0.1.4

Updating dependencies
Resolving dependencies... (0.6s)

  SolverProblemError

  The current project's Python requirement (>=3.7,<4.0) is not compatible with some of the required packages Python requirement:
    - kbackground requires Python >=3.7.1,<3.11, so it will not be satisfied for Python >=3.7,<3.7.1 || >=3.11,<4.0

  Because psfmachine depends on kbackground (0.1.4) which requires Python >=3.7.1,<3.11, version solving failed.

  at ~/.poetry/lib/poetry/puzzle/solver.py:241 in _solve
      237│             packages = result.packages
      238│         except OverrideNeeded as e:
      239│             return self.solve_in_compatibility_mode(e.overrides, use_latest=use_latest)
      240│         except SolveFailure as e:
    → 241│             raise SolverProblemError(e)
      242│
      243│         results = dict(
      244│             depth_first_search(
      245│                 PackageNode(self._package, packages), aggregate_package_nodes

  • Check your dependencies Python requirement: The Python requirement can be specified via the `python` or `markers` properties

    For kbackground, a possible solution would be to set the `python` property to ">=3.7.1,<3.11"

    https://python-poetry.org/docs/dependency-specification/#python-restricted-dependencies,
    https://python-poetry.org/docs/dependency-specification/#using-environment-markers
```

but `kbackground` complies with the suggestion.

Any idea?

jorgemarpa:
`python` and `kbackground` version issue fixed

christinahedges (commented):
Another excellent one I think this is ready to merge.

PR #47, opened 2022-02-03
:zap: tweaking PSFmachine to play with the K2 implementation

christinahedges:
I'm taking a look at PSFmachine for ruprecht, and making a few tweaks. I'm opening this issue to discuss these tweaks.

Here's a rough log of changes:

1. Changed the `_update_source_mask...` function so that it more aggressively clips out the faint edges of the PSF
2. Messed a little with some of the sane defaults for the masking function
3. Added a slightly more robust error catch for the "breaks" in time binning,
4. Get centroid takes a frame index so it will only calculate the centroid of the current frame unless "mean" is passed...(this might not be advisable)
5. `_get_source_mask` now puts the centroid to the centroid of the `frame_index`, making sure that per cadence PSFs are always centered at zero.
6. `build_frame_shape_model` recalculates the source mask at every step

The last point makes sure that the PSF is always centered at 0,0 when we fit it. This is important, as our model is only valid if the PSF it's modeling is centered.

However, the way it is currently set up, this becomes quite intractable for large datasets...!

### DIFF
--- a/src/psfmachine/machine.py
+++ b/src/psfmachine/machine.py
@@ -615,7 +615,7 @@ class Machine(object):
             ts = np.atleast_1d(frame_index)
         for t in ts:
             wgts = (
-                self.uncontaminated_source_mask.multiply(self.flux[t])
+                self.uncontaminated_source_mask.multiply(self.flux[t].copy())
                 .multiply(1 / self.source_flux_estimates[:, None])
                 .data
                 ** 2
@@ -1083,7 +1083,7 @@ class Machine(object):
         return fig
 
     def build_shape_model(
-        self, plot=False, flux_cut_off=1, frame_index="mean", **kwargs
+        self, plot=False, flux_atol=1, flux_rtol=0.001, frame_index="mean", **kwargs
     ):
         """
         Builds a sparse model matrix of shape nsources x npixels to be used when
@@ -1091,8 +1091,10 @@ class Machine(object):
 
         Parameters
         ----------
-        flux_cut_off: float
+        flux_atol: float
             the flux in COUNTS at which to stop evaluating the model!
+        flux_rtol: float
+            the relative flux of the PSF at which to stop evaluating the model
         frame_index : string or int
             The frame index used to build the shape model, if "mean" then use the
             mean value across time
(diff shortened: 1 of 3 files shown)


src/psfmachine/machine.py
```diff
@@ -323,12 +323,13 @@ def _create_delta_sparse_arrays(self, dist_lim=40, centroid_offset=[0, 0]):
 
     def _get_source_mask(
         self,
-        upper_radius_limit=28.0,
+        upper_radius_limit=36.0,
         lower_radius_limit=4.5,
         upper_flux_limit=2e5,
-        lower_flux_limit=100,
+        lower_flux_limit=40,
```
jorgemarpa, line 329: this makes the rough mask more generous, good when scene motion is considerable, and allows for fainter pixels

src/psfmachine/machine.py
```diff
@@ -599,22 +609,30 @@ def _get_centroids(self):
...
+            wgts = (
+                self.uncontaminated_source_mask.multiply(self.flux[t].copy())
+                .multiply(1 / self.source_flux_estimates[:, None])
+                .data
+                ** 2
+            )
             # mask out non finite values and background pixels
             k = (np.isfinite(wgts)) & (
-                self.uncontaminated_source_mask.multiply(self.flux[t]).data > 100
+                self.uncontaminated_source_mask.multiply(self.flux[t]).data > 10
```
jorgemarpa, line 625: I agree this value should be more generous. 
Let's add a comment that this value works well for Kepler/K2 background-subtracted pixels.

src/psfmachine/machine.py
```diff
@@ -323,12 +323,13 @@ def _create_delta_sparse_arrays(self, dist_lim=40, centroid_offset=[0, 0]):
...
         self,
-        upper_radius_limit=28.0,
+        upper_radius_limit=36.0,
         lower_radius_limit=4.5,
         upper_flux_limit=2e5,
-        lower_flux_limit=100,
+        lower_flux_limit=40,
         correct_centroid_offset=True,
         plot=False,
+        frame_index="mean",
```
jorgemarpa, line 332: add docstring for this argument.
Using `main` keeps the original behavior, right?

src/psfmachine/machine.py
```diff
@@ -1048,16 +1083,18 @@ def plot_time_model(self):
         return fig
 
     def build_shape_model(
-        self, plot=False, flux_cut_off=1, frame_index="mean", **kwargs
+        self, plot=False, flux_atol=1, flux_rtol=0.001, frame_index="mean", **kwargs
```
jorgemarpa, line 1086: I think these new arguments are better than the ones we had before.

jorgemarpa (commented):
I like the small tweaks you add to the API. Only a few of them change the default values, the rest only makes the API more robust. I don't think the first ones will make big changes for Kepler data. 

We need to update the tutorials to reflect the new keyword arguments we are introducing here. I saw the tutorials in `christinatweaks` branch aren't updated to the last version in master. 
Also, the change to `superstamp.py` will make the computing times in the `SSMachine` tutorials a bit longer. 

I will run the Tutorials as a way to check things are consistent with the master branch, I'll report back when it's done.

PR #50, opened 2022-02-10
Optional data binning in shape model

jorgemarpa:
This adds the option of binning the `(dx, dy)` data, taking the median value of `mean_f` if a certain number of bin members are found, or preserving data otherwise. 
This allows fitting a "cleaner" model where the PSF is bright.

Here a generic example:
Screen Shot 2022-02-09 at 13 12 14

and a `build_shape_model(bin_data=True)` example
image

The new `utils.threshold_binning()` function is early flexible. It allows to change the number of bins, the absolute threshold (number of bin members when taking the median), and use different statistics (e.g. median, mean). 
It takes ~200 ms for a set of 10k data points, that's the typical number of pixels in a ~200 TPF stack. For an FFI with ~100k pixels, it takes ~1.4 s.

### DIFF
--- a/src/psfmachine/machine.py
+++ b/src/psfmachine/machine.py
@@ -1077,12 +1077,6 @@ class Machine(object):
         # not contaminated etc
         self._get_uncontaminated_pixel_mask()
 
-        # remove pixels belonging to sources fainter than 1e4 (gaia flux)
-        # self.uncontaminated_source_mask = self.uncontaminated_source_mask.multiply(
-        #     (self.source_flux_estimates > 1e4)[:, None]
-        # )
-        # self.uncontaminated_source_mask.eliminate_zeros()
-
         # for iter in range(niters):
         flux_estimates = self.source_flux_estimates[:, None]
 
@@ -1119,10 +1113,9 @@ class Machine(object):
         # take value from Quantity is not necessary
         phi_b = self.uncontaminated_source_mask.multiply(self.phi).data
         r_b = self.uncontaminated_source_mask.multiply(self.r).data
-        print(r_b.shape)
 
         if bin_data:
-            nbins = 25 if mean_f.shape[0] <= 5e3 else 100
+            nbins = 30 if mean_f.shape[0] <= 5e3 else 100
             phi_b, r_b, mean_f = threshold_binning(
                 phi_b,
                 r_b,
@@ -1130,7 +1123,6 @@ class Machine(object):
                 bins=nbins,
                 abs_thresh=5,
             )
-        print(r_b.shape)
 
         # build a design matrix A with b-splines basis in radius and angle axis.
         A = _make_A_polar(
@@ -1316,7 +1308,7 @@ class Machine(object):
         dy = dy.data * u.deg.to(u.arcsecond)
 
         if bin_data:
-            nbins = 25 if mean_f.shape[0] <= 5e3 else 100
+            nbins = 30 if mean_f.shape[0] <= 5e3 else 100
             dx, dy, mean_f = threshold_binning(dx, dy, mean_f, bins=nbins, abs_thresh=5)
 
         fig, ax = plt.subplots(3, 2, figsize=(9, 10.5), constrained_layout=True)
--- a/src/psfmachine/machine.py
+++ b/src/psfmachine/machine.py
@@ -1115,8 +1115,8 @@ class Machine(object):
         r_b = self.uncontaminated_source_mask.multiply(self.r).data
 
         if bin_data:
-            nbins = 30 if mean_f.shape[0] <= 5e3 else 100
-            phi_b, r_b, mean_f, mean_f_err = threshold_bin(
+            nbins = 30 if mean_f.shape[0] <= 5e3 else 90
+            _, phi_b, r_b, mean_f, mean_f_err = threshold_bin(
                 phi_b,
                 r_b,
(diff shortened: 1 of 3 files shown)


src/psfmachine/machine.py
```diff
@@ -1113,6 +1114,11 @@ def build_shape_model(
         phi_b = self.uncontaminated_source_mask.multiply(self.phi).data
         r_b = self.uncontaminated_source_mask.multiply(self.r).data
 
+        if bin_data:
+            phi_b, r_b, mean_f = threshold_binning(
+                phi_b, r_b, mean_f, bins=100, abs_thresh=3
+            )
```
jorgemarpa: the number of bins and threshold need to be defined by the number of uncontaminated pixels.
These numbers I used here are good for an FFI with very dense PSF but don't work with sparse data from a TPF stack.

src/psfmachine/utils.py
```diff
@@ -443,3 +443,67 @@ def _combine_A(A, poscorr=None, time=None):
...
+        Binned X data.
+    new_y : numpy.array
+        Binned Y data.
+    new_f : numpy.array
+        Binned Z data.
+    """
+
+    if isinstance(bins, int):
+        bins = [bins, bins]
+    xedges = np.linspace(x.min(), x.max(), num=bins[0] + 1)
```
christinahedges:
```suggestion
    xedges = np.linspace(np.nanmin(x), np.nanmax(x), num=bins[0] + 1)
```

src/psfmachine/utils.py
```diff
@@ -443,3 +443,67 @@ def _combine_A(A, poscorr=None, time=None):
...
+    yedges = np.linspace(y.min(), y.max(), num=bins[1] + 1)
```
christinahedges:
```suggestion
    yedges = np.linspace(np.nanmin(y), np.nanmax(y), num=bins[1] + 1)
```

src/psfmachine/utils.py
```diff
@@ -443,3 +443,67 @@ def _combine_A(A, poscorr=None, time=None):
             format="csr",
         )
         return A2
+
+
+def threshold_binning(x, y, z, abs_thresh=10, bins=15, statistic=np.nanmedian):
```
christinahedges: `threshold_binning` might be better as a verb `threshold_bin`?

christinahedges (changes requested):
This is great! There are two things that are missing that I think we need before we can merge:

1. We need errors to be propagated. This is because we need to know how much weight to provide each bin when we fit it in the next step. We need to be able to pass errors on `x` and have them be aggregated in a reasonable way. We may want to use a weighted mean for the `statistic`, and use errors on the weighted mean in bins.

We could either pass in 'errors' or 'weights' on each point (we can brainstorm which makes sense).

2. We need some tests of the `threshold_binning` function. Here are a few sketches of what tests might look like

```python

def test_threshold_bin():
    # Pass integers
    x, y, z = 0, 0, 0
    threshold_bin(x, y, z)
    assert ...
    # Pass integer as list
    x, y, z = [0], [0], [0]
    threshold_bin(x, y, z)
    assert ...
    # Pass arrays that are too short
    x, y, z = np.arange(10), np.arange(10), np.ones(10)
    threshold_bin(x, y, z)
    assert ...
    # Pass a bunch of nans
    x, y, z = np.random.normal(size=1000), np.random.normal(size=1000), np.ones(1000) * np.nan
    threshold_bin(x, y, z)
    assert ...
    # Pass something that should actually pass
    x, y, z = np.random.normal(size=1000), np.random.normal(size=1000), np.random.normal(size=1000)
    threshold_bin(x, y, z)
    assert ...
    # Should be a check here that some points are binned, some points are not...

```

src/psfmachine/utils.py
```diff
@@ -443,3 +443,67 @@ def _combine_A(A, poscorr=None, time=None):
...
+    new_x : numpy.array
+        Binned X data.
+    new_y : numpy.array
+        Binned Y data.
+    new_f : numpy.array
+        Binned Z data.
+    """
+
+    if isinstance(bins, int):
+        bins = [bins, bins]
```
christinahedges, line 500: Maybe do some checks here to ensure that users aren't passing e.g. 1e10 bins

src/psfmachine/machine.py
```diff
@@ -1113,6 +1114,16 @@ def build_shape_model(
         phi_b = self.uncontaminated_source_mask.multiply(self.phi).data
         r_b = self.uncontaminated_source_mask.multiply(self.r).data
 
+        if bin_data:
+            nbins = 30 if mean_f.shape[0] <= 5e3 else 100
```
christinahedges: Should this be in cartesian coordinates? JMP and CH are discussing

jorgemarpa:
New version includes tests and unique function signature when providing or not errors for `z` values.

christinahedges (approved):
This looks GREAT to me!

src/psfmachine/machine.py
jorgemarpa: Set the `nbins` depending on the number of pixels available, for 3 typical ranges.

jorgemarpa:
@christinahedges I'm happy with the `nbins` depending on how many pixels are in the set. I commented on the explanation to keep track of the hardcoded numbers. 

I think is ready for merging...

PR #51, opened 2022-03-08
Source Max Mask

jorgemarpa:
This uses max (instead of mean) to get the source mask, this helps to include more pixels than taking the mean.

### DIFF
--- a/src/psfmachine/machine.py
+++ b/src/psfmachine/machine.py
@@ -366,9 +366,6 @@ class Machine(object):
 
         # The average flux, which we assume is a good estimate of the whole stack of images
         max_flux = np.nanmax(self.flux[self.time_mask], axis=0)
-        # max_flux_err = (self.flux_err[self.time_mask] ** 0.5).sum(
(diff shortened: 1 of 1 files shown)


christinahedges (approved):
This is a small, but vital change!! A few quick questions:
1. We definitely only need this change on one line? There are no other places we are taking the max?
2. I -think- it's right we don't want a keyword argument, we basically never want the mean, we want the max...
3. In that case, should we change `mean_flux` to `max_flux` to be explicit everywhere?

jorgemarpa:
> This is a small, but vital change!! A few quick questions:
> 
> 1. We definitely only need this change on one line? There are no other places we are taking the max?

This is the only place we take the max.

> 2. I -think- it's right we don't want a keyword argument, we basically never want the mean, we want the max...

I agree

> 3. In that case, should we change `mean_flux` to `max_flux` to be explicit everywhere?

Variable names updated

Ready to be merged

christinahedges (approved):

PR #52, opened 2022-03-15
Fitting time model for intra-quarter segments 

jorgemarpa:
This PR allows fitting different time models to light curve segments inside a quarter by using a common shape model and pixel lightcurve normalization.
It finds how many segments are in the quarter by finding discontinuities in the time data. Then iterates over these segments to build and fit time models accordingly.

The `build_time_model()` takes the `split_models` argument (boolean) to split or not by segments.
The `plot_time_model()` method takes the `segment` argument to select which segment to plot.

I fixed the downsampling in `_time_bin()` to capture the start/end of each segment.  This fix makes the one `pytest` fail. Need to see what's going on there.

TODO:
- [x] Maybe `split_models` argument could also accept an array of indexes for user-custom splits
- [x] Fix pytest fail for `_time_bin()`

### DIFF
--- a/src/psfmachine/machine.py
+++ b/src/psfmachine/machine.py
@@ -645,9 +645,11 @@ class Machine(object):
         """
 
         # Where there are break points in the time array
+        dts = np.diff(self.time)
         splits = np.append(
-            np.append(0, np.where(np.diff(self.time) > 0.1)[0] + 1), len(self.time)
+            np.append(0, np.where(dts > 5 * np.median(dts))[0] + 1), len(self.time)
         )
+        del dts
         # if using poscorr, find and add discontinuity in poscorr data
         if hasattr(self, "pos_corr1") and self.time_corrector in [
             "pos_corr",
@@ -681,10 +683,23 @@ class Machine(object):
         # within them
         breaks = []
         for spdx in range(len(splits_a)):
-            breaks.append(splits_a[spdx] + np.arange(0, dsplits[spdx]) * npoints)
-        # we include the last cadance as 99% of the sequence lenght
-        breaks.append(int(self.nt * 0.99))
-        breaks = np.hstack(breaks)
+            if not downsample:
+                breaks.append(splits_a[spdx] + np.arange(0, dsplits[spdx]) * npoints)
+            else:
+                # if downsample, the knots are spaced by npoints untill the end of the
+                # segment
+                breaks.append(
+                    np.linspace(
+                        splits_a[spdx],
+                        splits_b[spdx] - int(self.nt * 0.01),
+                        dsplits[spdx],
+                        dtype=int,
+                    )
+                )
+        breaks = np.unique(np.hstack(breaks))
+        # we include the last cadance as 99% of the sequence lenght if necessary
+        if breaks[-1] < self.nt - npoints:
+            breaks = np.append(breaks, int(self.nt * 0.99))
 
         if not downsample:
             # time averaged between breaks
@@ -722,13 +737,8 @@ class Machine(object):
                 ]
             )
         else:
-            dwns_idx = (
-                np.vstack(
-                    [np.median(t1) for t1 in np.array_split(np.arange(self.nt), breaks)]
-                )
-                .ravel()
-                .astype(int)
-            )
+            # use breaks as downsample points
+            dwns_idx = breaks
             tm = self.time[dwns_idx]
             ta = (self.time - tm.mean()) / (tm.max() - tm.mean())
             fm = np.asarray(
@@ -783,6 +793,12 @@ class Machine(object):
                 )
             pc1_smooth = np.array(pc1_smooth)
             pc2_smooth = np.array(pc2_smooth)
+            pc1_smooth = (pc1_smooth - pc1_smooth.mean()) / (
+                pc1_smooth.max() - pc1_smooth.mean()
+            )
+            pc2_smooth = (pc2_smooth - pc2_smooth.mean()) / (
+                pc2_smooth.max() - pc2_smooth.mean()
+            )
 
             # do poscorr binning
             if not downsample:
@@ -810,7 +826,7 @@ class Machine(object):
 
         return ta, tm, fm_raw, fm, fem
 
-    def build_time_model(self, plot=False, downsample=False):
+    def build_time_model(self, plot=False, downsample=False, split_models=False):
         """
         Builds a time model that moves the PRF model to account for the scene movement
         due to velocity aberration. It has two methods to choose from using the
@@ -827,6 +843,9 @@ class Machine(object):
         downsample: boolean
             If True the `time` and `pos_corr` arrays will be downsampled instead of
             binned.
+        split_models : boolean
+            Will split the light curve into segments to fit different time models with
+            a commong pixel normalization.
         """
         if hasattr(self, "pos_corr1") and self.time_corrector in [
             "pos_corr",
@@ -855,6 +874,7 @@ class Machine(object):
             ) = self._time_bin(npoints=self.n_time_points, downsample=downsample)
 
         self._whitened_time = time_original
+        self.downsample_time_knots = downsample
         # not necessary to take value from Quantity to do .multiply()
         dx, dy = (
             self.uncontaminated_source_mask.multiply(self.dra),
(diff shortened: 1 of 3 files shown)


src/psfmachine/machine.py
```diff
@@ -681,10 +683,23 @@ def _time_bin(self, npoints=200, downsample=False):
...
+                        splits_a[spdx],
+                        splits_b[spdx] - int(self.nt * 0.01),
+                        dsplits[spdx],
+                        dtype=int,
+                    )
+                )
+        breaks = np.unique(np.hstack(breaks))
+        # we include the last cadance as 99% of the sequence lenght if necessary
+        if breaks[-1] < self.nt - npoints:
+            breaks = np.append(breaks, int(self.nt * 0.99))
```
christinahedges: I'm not sure I understand what this is doing...

src/psfmachine/machine.py
```diff
@@ -783,6 +793,12 @@ def _time_bin(self, npoints=200, downsample=False):
                 )
             pc1_smooth = np.array(pc1_smooth)
             pc2_smooth = np.array(pc2_smooth)
+            pc1_smooth = (pc1_smooth - pc1_smooth.mean()) / (
```
christinahedges, line 793: Is this just normalization?

src/psfmachine/machine.py
```diff
@@ -827,6 +843,9 @@ def build_time_model(self, plot=False, downsample=False):
         downsample: boolean
             If True the `time` and `pos_corr` arrays will be downsampled instead of
             binned.
+        split_models : boolean
```
christinahedges: Maybe this could be something like `split_time_series`, `split_model` is harder to understand what is going on. But I don't really feel strongly.

src/psfmachine/machine.py
```diff
@@ -882,56 +902,108 @@ def build_time_model(self, plot=False, downsample=False):
...
         prior_sigma = np.ones(A3.shape[1]) * 10
         prior_mu = np.zeros(A3.shape[1])
 
-        for count in [0, 1, 2]:
-            sigma_w_inv = A3[k].T.dot(
-                (A3.multiply(1 / flux_err_binned.ravel()[:, None] ** 2)).tocsr()[k]
+        # fit the time model for each segment
+        # we find the splits in data
+        if split_models:
+            dts = np.diff(self.time)
```
christinahedges: This line is used in the beginning of the PR, should we have a method for doing this so that it's always the same, and easy to debug? Maybe in utils? Potentially overkill

```python
def get_breaks(time):
    dts = np.diff(self.time)
    return np.hstack([0, np.where(dts > 5 * np.median(dts))[0] + 1, len(self.time)])
```

src/psfmachine/machine.py
```diff
@@ -882,56 +902,108 @@ def build_time_model(self, plot=False, downsample=False):
...
+                time_binned[:, 0] < time_original[splits[bk + 1] - 1]
             )
-            B += prior_mu / (prior_sigma ** 2)
-            time_model_w = np.linalg.solve(sigma_w_inv, B)
-            res = flux_binned - A3.dot(time_model_w).reshape(flux_binned.shape)
-            res = np.ma.masked_array(res, (~k).reshape(flux_binned.shape))
-            bad_targets = sigma_clip(res, sigma=5).mask
-            bad_targets = (
-                np.ones(flux_binned.shape, bool) & bad_targets.any(axis=0)
+            print(seg_mask)
```
christinahedges: Don't want this in there!

src/psfmachine/machine.py
```diff
@@ -882,56 +902,108 @@ def build_time_model(self, plot=False, downsample=False):
...
+            if hasattr(self, "pos_corr1") and self.time_corrector in [
+                "pos_corr",
+                "centroid",
+            ]:
+                A3 = _combine_A(
+                    A2, poscorr=[poscorr1_binned[seg_mask], poscorr2_binned[seg_mask]]
+                )
+            else:
+                A3 = _combine_A(A2, time=time_binned[seg_mask])
+            print(A3.shape)
```
christinahedges: Print statement

src/psfmachine/machine.py
```diff
@@ -882,56 +902,108 @@ def build_time_model(self, plot=False, downsample=False):
...
+
+            # No saturated pixels
+            k = (
+                (flux_binned_raw < 1.4e5).any(axis=0)[None, :]
+                * np.ones(flux_binned_raw[seg_mask].shape, bool)
+            ).ravel()
+            # No faint pixels
+            k &= (
+                (flux_binned_raw[seg_mask] > 10).any(axis=0)[None, :]
```
christinahedges: I bet we could make that more aggressive. Shouldn't it be `all` and shouldn't the value be higher? 10 counts is hardly any, maybe all over 100 counts?

```python
k &= (
    (flux_binned_raw[seg_mask] > 100).all(axis=0)[None, :] 
    * np.ones(flux_binned_raw[seg_mask].shape, bool)
    ).ravel()
```

src/psfmachine/machine.py
```diff
@@ -882,56 +902,108 @@ def build_time_model(self, plot=False, downsample=False):
...
+                A3 = _combine_A(
+                    A2, poscorr=[poscorr1_binned[seg_mask], poscorr2_binned[seg_mask]]
+                )
+            else:
+                A3 = _combine_A(A2, time=time_binned[seg_mask])
+            print(A3.shape)
+
+            # No saturated pixels
+            k = (
+                (flux_binned_raw < 1.4e5).any(axis=0)[None, :]
```
christinahedges: Should this be `all` not `any`?

src/psfmachine/machine.py
```diff
@@ -882,56 +902,108 @@ def build_time_model(self, plot=False, downsample=False):
...
+                * np.ones(flux_binned_raw[seg_mask].shape, bool)
+            ).ravel()
+            # No faint pixels
+            k &= (
+                (flux_binned_raw[seg_mask] > 10).any(axis=0)[None, :]
+                * np.ones(flux_binned_raw[seg_mask].shape, bool)
+            ).ravel()
+            # No huge variability
+            k &= (
+                (np.abs(flux_binned[seg_mask] - 1) < 1).all(axis=0)[None, :]
```
christinahedges, line 951: This only gets rid of some of the variability. Most astrophysical variables will be less than this.....I think we can come up with a better mask.

src/psfmachine/machine.py
```diff
@@ -882,56 +902,108 @@ def build_time_model(self, plot=False, downsample=False):
...
-        self.time_model_w = time_model_w
-        self._time_masked = k
+            # we solve the segment by using `seg_mask` on all `*_binned` variables
+            for count in [0, 1, 2]:
+                sigma_w_inv = A3[k].T.dot(
+                    (
+                        A3.multiply(1 / flux_err_binned[seg_mask].ravel()[:, None] ** 2)
+                    ).tocsr()[k]
+                )
+                sigma_w_inv += np.diag(1 / prior_sigma ** 2)
```
christinahedges, line 966: This is where putting large errors for those "missing" pixels would work well to ensure those don't contribute (much) to the fit.

christinahedges (commented):
We're going to have to think about how to test this functionality...!

src/psfmachine/machine.py
jorgemarpa: Agree with the count limit. 
Yes, should be `all` to clip pixels that are always within the limits

src/psfmachine/machine.py
jorgemarpa: yes, only large-amplitude and period variability

src/psfmachine/machine.py
jorgemarpa: yes, that standardize the poscorrs to be zero-centered

src/psfmachine/machine.py
jorgemarpa: so it happens that when using the binning option, there's a knot missing at the end of the quarter (this doesn't happen when doing downsampling), so there I'm making sure that if know is missing within the last `npoints`, then I add one at the end. 

Here's an example if I do binning and I comment out lines 701 - 702. 
image

And here an example with downsampling
image

But maybe is not necessary

src/psfmachine/machine.py
jorgemarpa: oh yeah, `split_model` isn't the best word selection. Maybe something like `time_model_segments` or `split_time_model`.

src/psfmachine/machine.py
jorgemarpa: it's an overkill, but definitely easier to debug

jorgemarpa:
I fixed the `pytest` to take into account the normalization of the centroids/poscorrs.

src/psfmachine/machine.py
```diff
@@ -722,13 +731,8 @@ def _time_bin(self, npoints=200, downsample=False):
...
         else:
-            dwns_idx = (
-                np.vstack(
-                    [np.median(t1) for t1 in np.array_split(np.arange(self.nt), breaks)]
-                )
-                .ravel()
-                .astype(int)
-            )
+            # use breaks as downsample points
+            dwns_idx = breaks
```
christinahedges: Seems not necessary?

src/psfmachine/machine.py
```diff
@@ -882,56 +898,108 @@ def build_time_model(self, plot=False, downsample=False):
...
-            sigma_w_inv += np.diag(1 / prior_sigma ** 2)
-            # Fit the flux - 1
-            B = A3[k].T.dot(
-                ((flux_binned.ravel() - 1) / flux_err_binned.ravel() ** 2)[k]
+        # fit the time model for each segment
+        # use user input
+        if isinstance(split_time_model, (np.ndarray, list)):
+            # we make sure first and last index are included and sorted
+            splits = np.unique(np.append(split_time_model, [0, len(self.time)]))
+        else:
```
christinahedges:
```suggestion
        elif isinstance(split_time_model, bool):
```

christinahedges:
```suggestion
        elif isinstance(split_time_model, bool):
```

src/psfmachine/machine.py
```diff
@@ -882,56 +898,108 @@ def build_time_model(self, plot=False, downsample=False):
...
+            ]:
+                A3 = _combine_A(
+                    A2, poscorr=[poscorr1_binned[seg_mask], poscorr2_binned[seg_mask]]
+                )
+            else:
+                A3 = _combine_A(A2, time=time_binned[seg_mask])
+
+            # No saturated pixels
+            k = (
+                (flux_binned_raw < 1.4e5).all(axis=0)[None, :]
```
christinahedges: Hard coded value here...might be fine?

christinahedges: Hard coded value here...might be fine?

christinahedges: We can "inflate" this mask by ~2 pixel in the row dimension in order to ensure that we are not getting any ends of bleed columns. I think that would really help us in future. However, that might not belong in this PR!

christinahedges (commented):
This is really great functionality. I think this is done neatly, within the API we have. I do believe we could do with a refactor here in future. `A`, A_cp`, `A2`, and `A3` are getting confusing, and we have no way to build the perturbed model for a given time easily.  (see #55)



However, I think that's really for future work. I think this is what we need to get the functionality we need, and we'll look into a refactor in future.

PR #48, opened 2022-02-03
[WIP] :art: removing inplace centroids

christinahedges:

PR #54, opened 2022-03-18
Refactor API to remove in-place centroids, and update FFIMachine [do-NOT-merge]

christinahedges:
This PR refactors a lot of our API to make sure we have no in-place centroids. Here are some top changes:

1. We now have a `rough_mask`, which is our first pass at the masking
2. The `source_mask` is now made in a slightly more robust and clear way
3. FFIMachine now loads images in a more memory efficient way
4. We're no longer using that `rtol` to clip out faint parts of the PSF. We just use the `atol`, i.e. where the PSF has counts greater than some value.
5. For FFIMachine, we don't remove pixels that are saturated or "bad", we just make sure that in all the masks (`rough_mask`, `source_mask`, `uncontaminated_source_mask`) these are `False`.
6. I added a different background correction that uses splines to fit a smooth background model. This works well for TESS...I need to test it for Kepler FFIs
7. Makes FFIs work without removing pixels

This PR supersedes #48 

To Do

- [ ] Check this runs for Kepler FFIs!


christinahedges:
This explains what each of the three masks are:
image

christinahedges:
There's a notebook [here](https://gist.github.com/christinahedges/b3ea741bace80211658c09f56e774d66) which shows how to use this PR for getting the TESS photometry out...

src/psfmachine/machine.py
```diff
@@ -199,49 +213,95 @@ def shape(self):
...
+                    np.repeat(
+                        self.ra_centroid.value,
+                        self.dra.data.shape,
+                    ),
+                    (self.dra.nonzero()),
+                ),
+                shape=self.dra.shape,
+                dtype=float,
+            )
+            return self.dra - ra_offset
```
jorgemarpa, line 233: This is the core change to avoid in-place operation. Does this adds overhead when calling `self.dx`, I don't think it'd matter much though, only for the case of large sparse data.

src/psfmachine/machine.py
```diff
@@ -199,49 +213,95 @@ def shape(self):
...
+                shape=self.ddec.shape,
+                dtype=float,
+            )
+            return self.ddec - dec_offset
+
+    def _update_delta_arrays(self, frame_indices="mean"):
+        if self.nsources * self.npixels < 1e7:
+            self._update_delta_numpy_arrays()
+        else:
+            self._update_delta_sparse_arrays()
```
jorgemarpa, line 258: I think this should be renamed to just `_create_delta_*_arrays()` (or just `_delta_*_arrays()`) there's no "update" happening there. 
The `frame_indices="mean"` argument I think is unnecessary now.

src/psfmachine/machine.py
```diff
@@ -252,11 +312,11 @@ def _create_delta_sparse_arrays(self, dist_lim=40, centroid_offset=[0, 0]):
             Centroid offset for [ra, dec] to be included in dra and ddec computation.
             Default is [0, 0].
         """
+        if frame_indices == "mean":
+            frame_indices = np.where(self.time_mask)[0]
```
jorgemarpa, line 316: This isn't doing anything in the function.

src/psfmachine/machine.py
```diff
@@ -322,298 +366,104 @@ def _create_delta_sparse_arrays(self, dist_lim=40, centroid_offset=[0, 0]):
...
-        self,
-        upper_radius_limit=28.0,
-        lower_radius_limit=4.5,
-        upper_flux_limit=2e5,
-        lower_flux_limit=100,
-        correct_centroid_offset=True,
-        plot=False,
-    ):
-        """Find the pixel mask that identifies pixels with contributions from ANY NUMBER of Sources
+    def _get_source_mask(self, source_flux_limit=1):
```
jorgemarpa, line 369: this new version of `_get_source_mask` looks "simpler" than the original one, I mean, with less tunable params, which I like.

Is iterating 2 times good to converge into a solid `source_mask`?

jorgemarpa: A diagnostic plot for this one would be great, something to check we are capturing well the `f, r` dependency.

src/psfmachine/machine.py
```diff
@@ -322,298 +366,104 @@ def _create_delta_sparse_arrays(self, dist_lim=40, centroid_offset=[0, 0]):
...
-        dra_m = self.uncontaminated_source_mask.multiply(self.dra).data
-        ddec_m = self.uncontaminated_source_mask.multiply(self.ddec).data
-        for t in range(self.nt):
-            wgts = self.uncontaminated_source_mask.multiply(
-                np.sqrt(np.abs(self.flux[t]))
-            ).data
-            # mask out non finite values and background pixels
-            k = (np.isfinite(wgts)) & (
-                self.uncontaminated_source_mask.multiply(self.flux[t]).data > 100
+    def _get_centroid(self, plot=False):
```
jorgemarpa, line 435: we need short documentation here

jorgemarpa: Now, this method computes a single centroid value for all frames. Do we want to also have centroids in each frame?

src/psfmachine/machine.py
```diff
@@ -322,298 +366,104 @@ def _create_delta_sparse_arrays(self, dist_lim=40, centroid_offset=[0, 0]):
...
+        x, y = (
+            mask.multiply(self.dra).data,
+            mask.multiply(self.ddec).data,
+        )
+        c = (
+            mask.multiply(self.flux[self.time_mask].mean(axis=0))
+            .multiply(1 / self.source_flux_estimates[:, None])
+            .data
+        )
+        bm, nx, ny, nz, nze = threshold_bin(x, y, c, bins=20)
```
jorgemarpa, line 451: this is great use of the binning tool 😄

src/psfmachine/machine.py
```diff
@@ -1267,11 +1078,16 @@ def _get_mean_model(self):
...
+        # mean_model = mean_model.multiply(mean_model > relative_flux_limit)
+        # mean_model = mean_model.multiply(
+        #     mean_model.multiply(self.source_flux_estimates[:, None])
+        #     > absolute_flux_limit
+        # )
         mean_model.eliminate_zeros()
         self.mean_model = mean_model
 
-    def plot_shape_model(self, radius=20, frame_index="mean", bin_data=False):
+    def plot_shape_model(self, mask, bin_data=False, radius=None, frame_index="max"):
```
jorgemarpa, line 1090: I think `mask` should have a default value to easily work at the user level.

src/psfmachine/machine.py
```diff
@@ -1171,91 +1039,34 @@ def build_shape_model(
...
+        # self.uncontaminated_source_mask = _find_uncontaminated_pixels(self.source_mask)
+
+        #        self._update_source_mask_remove_bkg_pixels()
         # remove background pixels and recreate mean model
-        self._update_source_mask_remove_bkg_pixels(
-            flux_cut_off=flux_cut_off, frame_index=frame_index
-        )
+        # self._update_source_mask_remove_bkg_pixels(
+        #     flux_cut_off=flux_cut_off, frame_index=frame_index
+        # )
```
jorgemarpa, line 1050: why don't we need this now? Does the `mean_model` still have background pixels? before that was true because our original `source_mask` was circular, and then we did cleaning using the `shape_model`.

src/psfmachine/ffi.py
```diff
@@ -139,8 +141,11 @@ def __init__(
             # hardcoded to work for Kepler and TESS FFIs
             sparse_dist_lim=40 if meta["TELESCOP"] == "Kepler" else 210,
         )
-        self.meta = meta
-        self.wcs = wcs
+        self._mask_pixels()
+
+    @property
+    def flux_2d(self):
+        return self.flux.reshape((self.flux.shape[0], *self.image_shape))
```
jorgemarpa, line 148: this will make it easy for plotting stuff

src/psfmachine/ffi.py
```diff
@@ -455,7 +442,7 @@ def save_flux_values(self, output=None, format="fits"):
 
         return
 
-    def _remove_background(self, mask=None):
+    def _remove_background(self, mask=None, pixel_knot_spacing=10):
```
jorgemarpa, line 445: we can get rid of `photutils` dependency

src/psfmachine/ffi.py
```diff
@@ -105,18 +109,16 @@ def __init__(
...
+        thumb = np.min(self.flux, axis=0).reshape(self.image_shape)
+        gthumb = np.hypot(*np.gradient(thumb))
+        mask = (
+            ~sigma_clip(
+                np.ma.masked_array(gthumb, gthumb > 500),
+                sigma=3,
+                cenfunc=lambda x, axis: 0,
+            ).mask
+        ).ravel()
+        self._remove_background(mask=mask)
```
jorgemarpa, line 121: I think this can happens after `super.__init__` now.

src/psfmachine/ffi.py
```diff
@@ -583,16 +580,15 @@ def _mask_pixels(self, pixel_saturation_limit=1.2e5, magnitude_bright_limit=8):
...
-        return
+        pixel_mask = self.non_sat_pixel_mask & self.non_bright_source_mask
+        self.rough_mask = self.rough_mask.multiply(pixel_mask).tocsr()
+        self.rough_mask.eliminate_zeros()
+        self.source_mask = self.source_mask.multiply(pixel_mask).tocsr()
+        self.source_mask.eliminate_zeros()
+        self.uncontaminated_source_mask = self.uncontaminated_source_mask.multiply(
+            pixel_mask
+        ).tocsr()
+        self.uncontaminated_source_mask.eliminate_zeros()
```
jorgemarpa, line 591: this is great, never thought on including the sat/bright pixel mask into the `machine` mask, this way we can keep all original pixels and do nice image plots
😃

src/psfmachine/utils.py
```diff
@@ -402,7 +411,7 @@ def sparse_lessthan(arr, limit):
     return masked_arr
 
 
-def _combine_A(A, poscorr=None, time=None):
+def _combine_A(A, time, poscorr=None):
```
jorgemarpa, line 414: this one will disappear after merging with perturbation API

src/psfmachine/utils.py
```diff
@@ -542,3 +553,65 @@ def threshold_bin(x, y, z, z_err=None, abs_thresh=10, bins=15, statistic=np.nanm
         np.hstack(new_z),
         np.hstack(new_z_err),
     )
+
+
+def _find_uncontaminated_pixels(mask):
```
jorgemarpa, line 558: I'm not convinced this one belongs here, it used t be a hidden method of `machine.py`

jorgemarpa (commented):
In the near future, we need to merge the changes from #60 including the perturbation API. I think after that the only 'big' changes to `machine.py` are the way to compute the `source_mask` and `centroids`.

jorgemarpa:
I isolate the efficient FFI changes and open a new PR #71. We'll keep this PR open for future reference when including the in-place operations and new source mask methods.

Issue #56, opened 2022-03-26
Duplicated pixels are fully removed instead of keeping one instance


jorgemarpa:
I found the `TPFMachine` removes all duplicated pixels without even keeping one set of them, losing information.

Current behavior:
![image](https://user-images.githubusercontent.com/10449555/160218896-ab0ab056-85e2-4710-a954-23fed69963e1.png)
![image](https://user-images.githubusercontent.com/10449555/160218897-ebc38518-c95c-4420-acd6-163ee373e713.png)

Expected behavior:
![image](https://user-images.githubusercontent.com/10449555/160218784-4190804d-1482-47fc-9d7d-c32acfcaa3a5.png)
![image](https://user-images.githubusercontent.com/10449555/160218788-f11686df-129d-4230-bc75-153c3c3d6ec6.png)

PR #57, opened 2022-03-26
Fixing Issue #56 of duplicated pixels

jorgemarpa:
This PR fixes the duplicated pixel bug.
Here some examples of new behaviero:

image
image

Note: 
- I had to duplicate the masking steps to do nan & saturated pixel masking first, and then mask duplicated pixels.
- Pytest had to be updated to match the correct number of pixels. 
- I also included an extra test to check no duplicated pixels remain and that all valid pixels are kept only once after preprocess 

This solves #56

### DIFF
--- a/src/psfmachine/tpf.py
+++ b/src/psfmachine/tpf.py
@@ -1088,7 +1088,7 @@ def _parse_TPFs(tpfs, renormalize_tpf_bkg=True, **kwargs):
     sat_mask = []
     for tpf in tpfs:
         # Keplerish saturation limit
-        saturated = np.nanmax(tpf.flux, axis=0).T.value > 1.4e5
+        saturated = np.nanmax(tpf.flux, axis=0).T.value > 1.2e5
         saturated = np.hstack(
             (np.gradient(saturated.astype(float))[1] != 0) | saturated
         )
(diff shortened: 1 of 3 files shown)


jorgemarpa:
I included a couple of lines to add 3 columns before and after bleed columns. We can quickly change this number if we want later.

src/psfmachine/tpf.py
```diff
@@ -1088,10 +1088,11 @@ def _parse_TPFs(tpfs, renormalize_tpf_bkg=True, **kwargs):
     sat_mask = []
     for tpf in tpfs:
         # Keplerish saturation limit
-        saturated = np.nanmax(tpf.flux, axis=0).T.value > 1.4e5
-        saturated = np.hstack(
-            (np.gradient(saturated.astype(float))[1] != 0) | saturated
-        )
+        saturated = np.nanmax(tpf.flux, axis=0).T.value > 1.2e5
```
christinahedges, line 1091: This is hardcoded. I think when we switch this up for TESS it'll make things a pain. Let's leave it for now and make sure when we refactor we catch these

src/psfmachine/tpf.py
```diff
@@ -1088,10 +1088,11 @@ def _parse_TPFs(tpfs, renormalize_tpf_bkg=True, **kwargs):
...
+        # add 3 pixels before and after bleed column to be generous
+        for k in range(3):
```
christinahedges, line 1093: NICE

christinahedges (approved):
I think this looks good. My one comment is we could have a test of this saturated stuff by adding a saturated target to our test data. This might be hard to do, or might be adding a significant amount of data to our test data?

PR #59, opened 2022-03-30
:sparkles: Developing PerturbationMatrix class

christinahedges:
This PR is for developing a new PerturbationMatrix class which will ultimately factor our a lot of the headaches we're having inside machine

## To Do

- [x] Documentation (docstrings!)
- [ ] Merge into machine.py
- [ ] Write some tests for Kepler/K2/TESS data

### DIFF
--- a/src/psfmachine/perturbation.py
+++ b/src/psfmachine/perturbation.py
@@ -126,7 +126,10 @@ class PerturbationMatrix(object):
             nvec += 1
         if self.cbvs:
             nvec += self.ncbvs
-        s = nvec * (len(self.breaks) + 1)
+        if self.segments:
+            s = nvec * (len(self.breaks) + 1)
+        else:
+            s = nvec
         X = self.vectors[:, :s]
         w = np.linalg.solve(X.T.dot(X), X.T.dot(self.vectors[:, s:]))
         self.vectors[:, s:] -= X.dot(w)
@@ -134,7 +137,7 @@ class PerturbationMatrix(object):
 
     def plot(self):
         fig, ax = plt.subplots()
-ax.plot(self.time, self.vectors + np.arange(self.vectors.shape[1]) * 0.1)
+        ax.plot(self.time, self.vectors + np.arange(self.vectors.shape[1]) * 0.1)
         ax.set(xlabel="Time", ylabel="Vector", yticks=[], title="Vectors")
         return fig
 
--- a/src/psfmachine/perturbation.py
+++ b/src/psfmachine/perturbation.py
@@ -283,7 +283,7 @@ class PerturbationMatrix(object):
         return func
 
     def pca(self, y, ncomponents=5):
-        """In place operation to add the principal components of `y` to `other_vectors`
+        """In place operation to add the principal components of `y` to the design matrix
 
         Parameters
         ----------
--- a/src/psfmachine/perturbation.py
+++ b/src/psfmachine/perturbation.py
@@ -8,6 +8,7 @@ from psfmachine.utils import _make_A_cartesian
 import matplotlib.pyplot as plt
 from fbpca import pca
 from astropy.convolution import convolve, Box1DKernel
+from .utils import spline1d
 
 
 class PerturbationMatrix(object):
@@ -287,7 +288,7 @@ class PerturbationMatrix(object):
 
         return func
 
-    def pca(self, y, ncomponents=5, box_width=100):
+    def pca(self, y, ncomponents=5, long_time_scale=3, med_time_scale=0.5):
         """Adds the principal components of `y` to the design matrix
 
         Parameters
@@ -295,9 +296,14 @@ class PerturbationMatrix(object):
         y: np.ndarray
             Input flux array to take PCA of.
         """
-        return self._pca(y, ncomponents=ncomponents, box_width=box_width)
+        return self._pca(
+            y,
+            ncomponents=ncomponents,
+            long_time_scale=long_time_scale,
+            med_time_scale=med_time_scale,
+        )
 
-    def _pca(self, y, ncomponents=5, box_width=100):
+    def _pca(self, y, ncomponents=5, long_time_scale=3, med_time_scale=0.5):
         """This hidden method allows us to update the pca method for other classes"""
         if not y.ndim == 2:
             raise ValueError("Must pass a 2D `y`")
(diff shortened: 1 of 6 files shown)


src/psfmachine/perturbation.py
```diff
@@ -0,0 +1,333 @@
...
+            other_vectors=other_vectors,
+            poly_order=poly_order,
+            focus=focus,
+            cbvs=cbvs,
+            segments=segments,
+            clean=clean,
+        )
+
+    def __repr__(self):
+        return f"PerturbationMatrix3D"
```
jorgemarpa: the `f` is making flake8 not to pass

christinahedges:
Roughly how to use:

If `flux` and `flux_err` are arrays with dimensions (`ntimes` x `npixels`)

```python
P = PerturbationMatrix3d(time=time, dx=dx, dy=dy)
P.fit(flux, flux_err)
model = P.model()
```

`model` will then also have (`ntimes` x `npixels`)

src/psfmachine/perturbation.py
```diff
@@ -0,0 +1,364 @@
...
+            ]
+        )
+        focus *= np.asarray(
+            [
+                ((self.time - self.time[b]) >= 0).astype(float)
+                for b in np.hstack([0, self.breaks])
+            ]
+        )
+        self.vectors = np.hstack([self.vectors, np.nansum(focus, axis=0)[:, None]])
+        return
```
jorgemarpa: for the segments, focus values before the segment starts is actual `0` value, but after the break, the focus decays and never gets to `0` but to `e-32` values. I'll check if this affects somehow the result.

src/psfmachine/perturbation.py
```diff
@@ -0,0 +1,364 @@
...
+            nvec += self.ncbvs
+        s = nvec * (len(self.breaks) + 1)
+        X = self.vectors[:, :s]
+        w = np.linalg.solve(X.T.dot(X), X.T.dot(self.vectors[:, s:]))
+        self.vectors[:, s:] -= X.dot(w)
+        return
+
+    def plot(self):
+        fig, ax = plt.subplots()
+        ax.plot(self.time, self.vectors + np.arange(self.shape[1]) * 0.1, c="k")
```
jorgemarpa: this one crashes when calling from a `PerturbationMatrix3D` object because `self.shape` is different for `PerturbationMatrix3D` and `PerturbationMatrix`.

```suggestion
ax.plot(self.time, self.vectors + np.arange(self.vectors.shape[1]) * 0.1)
```

jorgemarpa: colors will make easy to separate components.

src/psfmachine/perturbation.py
```diff
@@ -0,0 +1,364 @@
...
+        points = []
+        b = np.hstack([0, self.breaks, len(self.time) - 1])
+        for b1, b2 in zip(b[:-1], b[1:]):
+            p = np.arange(b1, b2, self.resolution)
+            if p[-1] != b2:
+                p = np.hstack([p, b2])
+            points.append(p)
+        points = np.unique(np.hstack(points))
+
+        def func(x, quad=False):
```
jorgemarpa, line 253: I assume `quad` is here to keep the same signature as in `bindown`

src/psfmachine/perturbation.py
```diff
@@ -0,0 +1,364 @@
...
+
+        return func
+
+    def _get_bindown_func(self):
+        b = np.hstack([0, self.breaks, len(self.time) - 1])
+        points = np.hstack(
+            [np.arange(b1, b2, self.resolution) for b1, b2 in zip(b[:-1], b[1:])]
+        )
+        points = points[~np.in1d(points, np.hstack([0, len(self.time) - 1]))]
+        points = np.sort(np.unique(np.hstack([points, self.breaks])))
```
jorgemarpa: `np.unique` returns sorted values.

src/psfmachine/perturbation.py
```diff
@@ -0,0 +1,364 @@
...
+                            self.cartesian_matrix * vector[time_index],
+                            format="csr",
+                        )
+                        for vector in (self.vectors).T
+                    ],
+                    format="csr",
+                ).dot(self.weights)
+                for time_index in time_indices
+            ]
+        )
```
jorgemarpa: I tested this and it seems to run super slow, about 1.7 s/iter.
Also if we create a `PerturbationMatrix3D` object with the `dx, dy` uncontaminated pixels, then, we'll need a 2nd `PerturbationMatrix3D` with the pixels in `source_mask`. So I think we need a `pixel_mask` argument at initialization to take care of this.

tests/test_perturbation.py
```diff
@@ -0,0 +1,120 @@
...
+        )
+        p = PerturbationMatrix(time=time, other_vectors=1, focus=False)
+    p = PerturbationMatrix(
+        time=time, other_vectors=np.random.normal(size=(2, 100)), focus=False
+    )
+    p = PerturbationMatrix(
+        time=time, other_vectors=np.random.normal(size=(100, 2)), focus=False
+    )
+    assert p.vectors.shape == (100, 6)
+    time = np.hstack([np.arange(0, 10, 0.1), np.arange(15, 25, 0.1)])
```
jorgemarpa, line 27: With this time array we can also test the `segment` mode

src/psfmachine/perturbation.py
```diff
@@ -0,0 +1,364 @@
...
+
+    def plot_model(self, time_index=0):
+        if not hasattr(self, "weights"):
+            raise ValueError("Run `fit` first.")
+        fig, ax = plt.subplots()
+        ax.scatter(self.dx, self.dy, c=self.model(time_index)[0])
```
jorgemarpa, line 564: I think you meant to check if `self.model`  exists

jorgemarpa (commented):
This is great! I think the basic functionalities are implemented here.

Something that will be good to have is a `pixel_mask` argument in `PerturbationMatrix3D` so we can create `matrix` with the `source_mask` (for fitting) pixels and the `uncontaminated_source_mask` (for building) in one perturbation object. 

I did a couple of "performance" test with TPF data and turned out that the `self.model` method runs really slow, about 1.7 sec/iter. We need to find why this is so slow.

src/psfmachine/perturbation.py
```diff
@@ -0,0 +1,364 @@
...
+        return
+
+    def _clean_vectors(self):
+        """Remove time polynomial from other vectors"""
+        nvec = self.poly_order + 1
+        if self.focus:
+            nvec += 1
+        if self.cbvs:
+            nvec += self.ncbvs
+        s = nvec * (len(self.breaks) + 1)
```
jorgemarpa: This will change the component index even when no segments are used.
Example, if `segments=False` and `other_vectors=[x_cen, y_cen, x_cen * y_cen]`, then `s` will be 12 (for a 3 segment case). Then the index is wrong, then it won't remove the time component to the centroids.

src/psfmachine/perturbation.py
christinahedges: I know....I really wanted to be explicit for the reader since these have to be sorted....but we can omit this

christinahedges:
```suggestion
        points = np.unique(np.hstack([points, self.breaks]))
```

src/psfmachine/perturbation.py
christinahedges: I added a line which will nix this component to zero when it dips below 1e-10

christinahedges:
I added a PCA method so you can do something like:

```python
y = [whitened_timeseries, whitened_timeseries]
p = PerturbationMatrix(...)
p.pca(y, ncomponents=10)
p.fit(y)
```

src/psfmachine/perturbation.py
```diff
@@ -0,0 +1,478 @@
...
+    def __init__(
+        self,
+        time: npt.ArrayLike,
+        other_vectors: Optional[list] = None,
+        poly_order: int = 3,
+        focus: bool = False,
+        segments: bool = True,
+        resolution: int = 10,
+        bin_method: str = "bin",
+        focus_exptime=50,
```
jorgemarpa: this argument needs to be visible from `PerturbationMatrix3D`

src/psfmachine/perturbation.py
```diff
@@ -0,0 +1,478 @@
...
+        for count in range(3):
+            U, s, V = pca(y[:, k], ncomponents, n_iter=30)
+            k[k] &= (np.abs(V) < 0.5).all(axis=0)
+
+        self._vectors = np.hstack([self._vectors, U])
+        if self.segments:
+            self._vectors = self._cut_segments(self._vectors)
+        self._validate()
+        self._clean_vectors()
+        self.matrix = sparse.csr_matrix(self.bin_func(self.vectors))
```
jorgemarpa: @christinahedges  I found the bug when doing `self.model()`. We are not updating `self._cartesian_stacked`  to reflect the addition of the new PCA components to the `self.vectors`. That's why the inconsistent shapes.

Actually, when we do `self.pca()` on a `PerturbationMatrix3D` object we are changing the attributes of a `PerturbationMatrix` object that was already initialized with no PCA components. Therefore everything that happens after the `super().__init__` in `PerturbationMatrix3D` does not get updated.

jorgemarpa (commented):
@christinahedges  I found the bug with the inconsistent shape when doing `self.model()`. The problem is when using `PerturbationMatrix` through `PerturbationMatrix3D` that some attributes from the latter are not updated to reflect the addition of PCA components in `self.vectors`. See my comments for more details.

christinahedges:
Any vectors that aren't the time polynomial are now set to zero mean in each "segment" during `_clean_vectors` method. The `pca` method can now be repeated and it will just update the pca components, not append ad infinitum

src/psfmachine/perturbation.py
christinahedges: This was a problem because I only tested the `PerturbationMatrix` class with pca components, not the `PerturbationMatrix3d`!! Lesson learned, this now has a fix.

src/psfmachine/perturbation.py
```diff
@@ -0,0 +1,500 @@
...
+
+    def pca(self, y, ncomponents=5):
+        """Adds the principal components of `y` to the design matrix
+
+        Parameters
+        ----------
+        y: np.ndarray
+            Input flux array to take PCA of.
+        """
+        self._pca(y, ncomponents=5)
```
jorgemarpa:
```suggestion
        self._pca(y, ncomponents=ncomponents)
```

src/psfmachine/perturbation.py
```diff
@@ -0,0 +1,576 @@
...
+            Input flux array to take PCA of.
+        ncomponents: int
+            Number of principal components to use
+        long_time_scale: float
+            The time scale where variability is considered "long" term. Should be
+            in the same units as `self.time`.
+        med_time_scale: float
+            The time scale where variability is considered "medium" term. Should be
+            in the same units as `self.time`. Variability longer than `long_time_scale`,
+            or shorter than `med_time_scale`, will be removed before building components
```
jorgemarpa, line 316:
```suggestion
smooth_time_scale: float
            Amount to smooth the components, using a spline in time.
             If 0, the components will not be smoothed.
```

src/psfmachine/perturbation.py
```diff
@@ -0,0 +1,576 @@
...
+                            ]
+                        )
+                        ** 0.5
+                    )
+            else:
+                raise ValueError("Wrong size to bin")
+
+        return func
+
+    def pca(self, y, ncomponents=5, smooth_time_scale=0):
```
jorgemarpa, line 298: to be consistent with the rest

```suggestion
    def pca(self, y, ncomponents=3, smooth_time_scale=0):
```

src/psfmachine/perturbation.py
```diff
@@ -0,0 +1,576 @@
...
+        """Adds the principal components of `y` to the design matrix
+
+        Will add two time scales of principal components, definied by `long_time_scale`
+        and `med_time_scale`.
```
jorgemarpa:
```suggestion
```

src/psfmachine/perturbation.py
jorgemarpa: What is the scale of `smooth_time_scale`? 

```suggestion
        smooth_time_scale: float
             Amount to smooth the components, using a spline in time.
             If 0, the components will not be smoothed.
```

src/psfmachine/perturbation.py
```diff
@@ -0,0 +1,576 @@
...
+    #
+    #     if s != self.vectors.shape[1]:
+    #         X = self.vectors[:, :s]
+    #         w = np.linalg.solve(X.T.dot(X), X.T.dot(self.vectors[:, s:]))
+    #         self.vectors[:, s:] -= X.dot(w)
+    #         # Each segment has mean zero
+    #         self.vectors[:, s:] -= np.asarray(
+    #             [v[v != 0].mean() * (v != 0) for v in self.vectors[:, s:].T]
+    #         ).T
+    #     return
```
jorgemarpa, line 154: maybe we can still keep this function in case we want to remove the temporal trend for other components. But it isn't necessary with the current API.

jorgemarpa (commented):
This looks great! I tested a handful of component combinations and other parameters to check the matrices look Ok. 
It fulfills what we always thought about the perturbation matrix API.

I just added a couple of comments regarding docstrings. Besides those, I think this is ready to be merged.

PR #60, opened 2022-04-06
Machine uses perturbation class

jorgemarpa:
This PR adapts `machine` to use the new `perturbation` classes introduced in #59.

It modifies the following methods:
- `self.build_time_model()`
- `self.plot_time_model()`
- `self.fit_models()`
- new `self._get_perturbed_model()`, computes the perturbed matrix for a given time.

It adds a new attribute `self.P3` which is a `PerturbationMatrix3D` object that manages the perturbation matrix.

TODO:
- [x] fix pytest
- [x] fix flake8
- [x] test focus component
- [x] test centroid components
- [x] check for attribute consistency

EDIT:
The original `mac.time_corrector = "pos_corr", "centroid"` options can be mimic by including the following arguments to `PerturbationMatrix3D` at initialization: `other_vectors = [poscorr1, poscorr2, poscorr1 * poscorr2], poly_order=0`. In this way the perturbation matrix will use only the centroids/poscorr as components with no time.


src/psfmachine/machine.py
```diff
@@ -837,289 +635,197 @@ def build_time_model(self, plot=False, downsample=False, split_time_model=False)
...
+        uncontaminated_pixels = uncontaminated_pixels.data
 
-        A_c = _make_A_cartesian(
-            dx,
-            dy,
-            n_knots=self.n_time_knots,
-            radius=self.time_radius,
-            spacing=self.cartesian_knot_spacing,
+        # add other vectors if asked, centroids or poscorrs
+        splits = get_breaks(self.time)
```
christinahedges: not necessary anymore, we should be doing this inside perturbation matrix. I think you're doing this because you're going to clean up the centroids, but I think we should also do that inside `PerturbationMatrix`

src/psfmachine/machine.py
```diff
@@ -837,289 +635,197 @@ def build_time_model(self, plot=False, downsample=False, split_time_model=False)
...
         )
-        A2 = sparse.vstack([A_c] * time_binned.shape[0], format="csr")
-        # Cartesian spline with time dependence
-        if hasattr(self, "pos_corr1") and self.time_corrector in [
-            "pos_corr",
-            "centroid",
-        ]:
-            # Cartesian spline with poscor dependence
-            A3 = _combine_A(A2, poscorr=[poscorr1_binned, poscorr2_binned])
+        if self.time_corrector == "polynomial":
```
christinahedges: maybe we want to think about this API since polynomial will always be the default, and in fact we can't turn it off.

src/psfmachine/machine.py
```diff
@@ -837,289 +635,197 @@ def build_time_model(self, plot=False, downsample=False, split_time_model=False)
...
+            if self.time_corrector == "pos_corr" and hasattr(self, "pos_corr1"):
+                mpc1 = np.nanmedian(self.pos_corr1, axis=0)
+                mpc2 = np.nanmedian(self.pos_corr2, axis=0)
+                # other_vectors =
+            elif self.time_corrector == "centroid" and hasattr(self, "ra_centroid"):
+                mpc1 = self.ra_centroid.to("arcsec").value / 4
+                mpc2 = self.dec_centroid.to("arcsec").value / 4
+
+            # smooth the vectors
+            mpc1_smooth, mpc2_smooth = smooth_vector(
```
christinahedges: We shouldn't use a Gaussian smooth here we know it's noisy, we should replace it with a spline fit. I think we can have that be part of the 'cleaning' in perturbation matrix

src/psfmachine/machine.py
```diff
@@ -837,289 +635,197 @@ def build_time_model(self, plot=False, downsample=False, split_time_model=False)
...
+            )
+            # combine them as the first order
+            other_vectors = [mpc1_smooth, mpc2_smooth, mpc1_smooth * mpc2_smooth]
+
+        # create a 3D perturbation matrix
+        P3 = PerturbationMatrix3D(
+            time=_whitened_time,
+            dx=dx,
+            dy=dy,
+            segments=split_time_segments,
```
christinahedges: let's keep these keywords the same across `machine` and `perturbation` otherwise it will get confusing. I don't mind if we change some `PerturbationMatrix` keywords, but we should definitely try to be consistent.

src/psfmachine/machine.py
```diff
@@ -837,289 +635,197 @@ def build_time_model(self, plot=False, downsample=False, split_time_model=False)
...
-                        (flux_binned[seg_mask].ravel() - 1)
-                        / flux_err_binned[seg_mask].ravel() ** 2
-                    )[k]
-                )
-                B += prior_mu / (prior_sigma ** 2)
-                time_model_w = np.linalg.solve(sigma_w_inv, B)
-                res = flux_binned[seg_mask] - A3.dot(time_model_w).reshape(
-                    flux_binned[seg_mask].shape
+        # get uncontaminated pixel norm-flux
+        flux, flux_err = np.array(
```
christinahedges, line 717: This should be the uncontaminated flux mask, not source flux mask

src/psfmachine/machine.py
```diff
@@ -837,289 +635,197 @@ def build_time_model(self, plot=False, downsample=False, split_time_model=False)
...
-        self.time_model_w = np.asarray(seg_time_model_w)
-        self._time_masked = np.asarray(pix_mask_k, dtype=object)
+                for idx in range(self.nt)
+            ]
+        ).transpose((1, 0, 2))
+        flux_norm = flux / np.nanmean(flux, axis=0)
+        flux_err_norm = flux_err / np.nanmean(flux, axis=0)
+
+        # bindown flux arrays
+        flux_binned_raw = P3.bin_func(flux)
```
christinahedges: Why are we working in the binned space to identify bad pixels? It seems like we could do that in the unbinned space?

src/psfmachine/machine.py
```diff
@@ -837,289 +635,197 @@ def build_time_model(self, plot=False, downsample=False, split_time_model=False)
...
+        flux_binned = P3.bin_func(flux_norm)
+        flux_err_binned = P3.bin_func(flux_err_norm)
+
+        # No saturated pixels, 1e5 is a hardcoded value for Kepler.
+        k = (flux_binned_raw < 1e5).all(axis=0)[None, :] * np.ones(
```
christinahedges: Do we already have a saturated pixel mask somewhere? Can we reuse it here? That will let us generously remove pixels

src/psfmachine/machine.py
```diff
@@ -837,289 +635,197 @@ def build_time_model(self, plot=False, downsample=False, split_time_model=False)
...
+            flux_binned_raw.shape, bool
+        )
+        # No faint pixels, 100 is a hardcoded value for Kepler.
+        k &= (flux_binned_raw > 100).all(axis=0)[None, :] * np.ones(
```
christinahedges: We should take this hard coded value and put it as a keyword argument, so that we can change it for TESS

src/psfmachine/machine.py
```diff
@@ -837,289 +635,197 @@ def build_time_model(self, plot=False, downsample=False, split_time_model=False)
...
+            flux_binned_raw.shape, bool
+        )
+        # No huge variability
+        k &= (np.abs(flux_binned - 1) < 1).all(axis=0)[None, :] * np.ones(
```
christinahedges: This seems like a bad test. We should make this better

src/psfmachine/machine.py
```diff
@@ -837,289 +635,197 @@ def build_time_model(self, plot=False, downsample=False, split_time_model=False)
...
+            flux_binned.shape, bool
+        )
+        # No nans
+        k &= np.isfinite(flux_binned) & np.isfinite(flux_err_binned)
+        k = np.all(k, axis=0)
+        # combine good-behaved pixels with uncontaminated pixels
+        k &= uncontaminated_pixels
```
christinahedges, line 748: You can get rid of this by just using the `uncontaminated_source_mask` above, I think?

src/psfmachine/machine.py
```diff
@@ -837,289 +635,197 @@ def build_time_model(self, plot=False, downsample=False, split_time_model=False)
...
+
+        # iterate to remvoe outliers
+        for count in [0, 1, 2]:
+            # need to add pixel_mask=k
+            P3.fit(flux_norm, flux_err=flux_err_norm, pixel_mask=k)
+            res = flux_binned - P3.matrix.dot(P3.weights).reshape(flux_binned.shape)
+            # res = np.ma.masked_array(res, (~k).reshape(flux_binned.shape))
+            res = np.ma.masked_array(res, ~np.tile(k, flux_binned.shape[0]).T)
+            bad_targets = sigma_clip(res, sigma=5).mask
```
christinahedges: Does this need an `axis` keyword argument? I think it does. Maybe calculate the reduced chi squared for each time series, and then sigma_clip on that:
Something like:
```python
chi2 = np.sum((flux_binned - model)**2/(flux_err_binned**2), axis=0)/nbins
bad_targets = sigma_clip(chi2, sigma=5).mask
```

(You might have to double check my axis)

src/psfmachine/machine.py
```diff
@@ -837,289 +635,197 @@ def build_time_model(self, plot=False, downsample=False, split_time_model=False)
...
+                for idx in range(self.nt)
+            ]
+        ).transpose((1, 0, 2))
+        flux_norm = flux / np.nanmean(flux, axis=0)
+        flux_err_norm = flux_err / np.nanmean(flux, axis=0)
+
+        # bindown flux arrays
+        flux_binned_raw = P3.bin_func(flux)
+        flux_binned = P3.bin_func(flux_norm)
+        flux_err_binned = P3.bin_func(flux_err_norm)
```
christinahedges: Make sure to use the `quad` keyword here otherwise it will not bin correctly!
```suggestion
        flux_err_binned = P3.bin_func(flux_err_norm, quad=True)
```

src/psfmachine/machine.py
```diff
@@ -837,289 +635,197 @@ def build_time_model(self, plot=False, downsample=False, split_time_model=False)
...
+            P3.fit(flux_norm, flux_err=flux_err_norm, pixel_mask=k)
+            res = flux_binned - P3.matrix.dot(P3.weights).reshape(flux_binned.shape)
+            # res = np.ma.masked_array(res, (~k).reshape(flux_binned.shape))
+            res = np.ma.masked_array(res, ~np.tile(k, flux_binned.shape[0]).T)
+            bad_targets = sigma_clip(res, sigma=5).mask
+            bad_targets = bad_targets.any(axis=0)
+            k &= ~bad_targets
+
+        # bookkeeping
+        self.flux_binned = flux_binned
```
christinahedges, line 774: Is this necessary? I think you probably are keeping it for plotting

src/psfmachine/machine.py
```diff
@@ -837,289 +635,197 @@ def build_time_model(self, plot=False, downsample=False, split_time_model=False)
...
+        self._time_masked_pix = k
+        self.P3 = P3
```
christinahedges: Can we call this `P` instead of `P3`? I think it's easier to read it that way as a ⭐ matrix ⭐

src/psfmachine/machine.py
```diff
@@ -837,289 +635,197 @@ def build_time_model(self, plot=False, downsample=False, split_time_model=False)
...
+        time_indices = np.atleast_1d(time_indices)
+        if isinstance(time_indices[0], bool):
+            time_indices = np.where(time_indices[0])[0]
+
+        perturbed_model = []
+        for tdx in time_indices:
+            X = self.mean_model.copy()
+            X.data *= self.P3.model(time_indices=tdx).ravel()
+            perturbed_model.append(X)
+        self.perturbed_model = perturbed_model
```
christinahedges:
```suggestion
        self.perturbed_model = np.asarray(perturbed_model)
```

christinahedges: I don't think this is quite the API we want, do we want something like this:

```python
mac.mean_model
mac.perturbed_model(time_index)
```

because then we can do

```python
mac.perturbed_model(time_index).dot(mac.psf_weights)
```

Right now you're setting an attribute (`self.perturbed_model`) and that's going to waste a lot of memory, and we can't do the dot product above. I think we just want a class method that returns a matrix. Using your code above:

```python

   def perturbed_model(self, time_index=0):
        """
        Computes the perturbed model at given times
        """        
        X = self.mean_model.copy()
        X.data *= self.P3.model(time_indices=time_index).ravel()
        return X
```

What exactly is that ravel doing?

src/psfmachine/machine.py
```diff
@@ -837,289 +635,197 @@ def build_time_model(self, plot=False, downsample=False, split_time_model=False)
...
 
-        Parameters
-        ----------
-        segment : int
-            Which light curve segment will be plotted, default is first one.
+        return
+
+        # raise NotImplementedError
+
+    def plot_time_model(self):
```
christinahedges, line 789: This is a great plot. I don't think we really want it in the binned space, even though I know that's where we fit it. I think we want to be able to see each frame.

```python
mac.plot_time_model(time_index=0)
```

That will let us see if, for example, the first frame is really bad because of focus change.

src/psfmachine/machine.py
```diff
@@ -837,289 +635,197 @@ def build_time_model(self, plot=False, downsample=False, split_time_model=False)
...
+        model_binned = self.P3.matrix.dot(self.P3.weights).reshape(
+            self.flux_binned.shape
         )
+        fig, ax = plt.subplots(2, 2, figsize=(9, 7), facecolor="w")
+        # k1 = self._time_masked_pix.reshape(self.flux_binned.shape)[0].astype(bool)
         im = ax[0, 0].scatter(
-            dx[k1],
-            dy[k1],
-            c=flux_binned[seg_mask][0][k1],
+            self.P3.dx[self._time_masked_pix],
```
christinahedges: Let's use `self.dx`

Those should be the same, right?

src/psfmachine/machine.py
```diff
@@ -1553,89 +1259,39 @@ def fit_model(self, fit_va=False):
...
-                        (tdx >= self.seg_splits[k]) & (tdx < self.seg_splits[k + 1])
-                        for k in range(len(self.seg_splits) - 1)
-                    ]
-                )[0]
-                X.data *= (
-                    A_cp3.multiply(t_mult).dot(self.time_model_w[seg_num].ravel()) + 1
-                )
-                X = X.T
+
+                X.data *= self.P3.model(time_indices=tdx).ravel()
```
christinahedges:
```suggestion
                X = self.perturbed_model(tdx)
```

src/psfmachine/machine.py
jorgemarpa: I use this to smooth out the poscor/centroids without flattening the breaks in the data.

jorgemarpa: I'll move this into `utils.smooth_vector` and include the `time` as an optional argument to allow for discontinuity awareness.

src/psfmachine/machine.py
jorgemarpa: Ok, I agree.

src/psfmachine/machine.py
jorgemarpa: I think the `other_vectors` should be already smooth before being passed to `perturbation` because we could pass other types of vectors that don't need "smoothing" or "cleaning". I think the "cleaning" inside `perturbation` only attends to removing the degeneracy in the basis we are using, aka removing the time trend.

jorgemarpa: I'll add an auxiliary function in `utils` to apply a spline smoothing to a vector vector `y`.

jorgemarpa: Here's an example using the new `spline_smooth()` function in `utils.py`:

image

src/psfmachine/machine.py
jorgemarpa: we do have a saturated pixel removal section, but that happens before initializing `machine` and effectively deleting pixels. Here we are just masking out "too bright" pixels to fit the time model. Also here the limit is slightly lower (`1e5` vs `1.2e5` in `tpf.py`).

src/psfmachine/machine.py
jorgemarpa: yes!

src/psfmachine/machine.py
jorgemarpa: I see what you mean, but we're using `flux, flux_err` to mask good-behaved pixels to then use that mask `k` in `P3` which was created with `source_mask` number of pixels instead of uncontaminated pixels.

src/psfmachine/machine.py
jorgemarpa: This is related to your comment above for line 703.
If I do that, then the shape of `k` won't match the number of pixels passed to the perturbation object.

src/psfmachine/machine.py
jorgemarpa: I improved the `spline_smooth` function. it should work with any type of vector now defining the number of knots and breaks if wanted.

src/psfmachine/machine.py
jorgemarpa: we don't have that attribute in `machine`

src/psfmachine/machine.py
jorgemarpa: That makes sense. 
The `.ravel()` is because `P3.model()` returns a column vector `(1, n_pix)`.

src/psfmachine/machine.py
jorgemarpa: .

src/psfmachine/machine.py
jorgemarpa: yes, that's for plotting in `plot_time_model()` but if we plot a frame instead, then it is not necessary.

src/psfmachine/machine.py
jorgemarpa: yes. I simplified this to have the same way as adding focus or PCA components.

pyproject.toml
```diff
@@ -36,7 +36,8 @@ photutils = "^1.1.0"
 diskcache = "^5.3.0"
 imageio = "^2.9.0"
 imageio-ffmpeg = "^0.4.5"
-kbackground = "^0.1.7"
+kbackground = "0.1.7"
```
christinahedges:
```suggestion
kbackground = "^0.1.7"
```

jorgemarpa (commented):
@christinahedges I reviewed this PR again, added some minor docstrings, and removed duplicated functions in `utils.py`. 

It looks Ok to me. Please review again before merging

src/psfmachine/perturbation.py
```diff
@@ -298,16 +308,18 @@ def func(x, quad=False):
     def pca(self, y, ncomponents=5, smooth_time_scale=0):
         """Adds the principal components of `y` to the design matrix
 
+        Will add two time scales of principal components, definied by `long_time_scale`
```
christinahedges: This isn't true anymore can we fix this docstring here?

src/psfmachine/tpf.py
```diff
@@ -143,6 +144,22 @@ def remove_background_model(self, plot=False, data_augment=None):
         # keep track of which pixels comes from TPFs
         pixels_in_tpf = np.ones_like(self.row, dtype=bool)
 
+        # enheance pixel mask
+        time_corr = np.nanpercentile(bkg_flux, 20, axis=1)[:, None]
```
christinahedges, line 152: Why is this 20%? Is this why stuff goes negative so often? Can you add a short docstring explainer about what's going on here?

christinahedges (commented):
I think this looks good, just two very minor docstring changes and then I will merge

src/psfmachine/tpf.py
jorgemarpa: That part is only used to create a naive model of the background as a function of time to then find and mask out variable pixels. So it does not play into building the bkg model. 
Changing the percentile will only change which pixels to look at in the distribution each time.

PR #63, opened 2022-06-22
Option to use CBVs and better priors 

jorgemarpa:
Two small changes/new features:

- `machine.build_time_model()` accepts `other_vectors` as argument to add other type of model components from user-level, e.g. using CBVs as regressors.
- Improved priors for in `PerturbationMatrix`. The prior follows ~ N(1, 0,5) which enforces "small" deviations from the mean model. This accounts for the mismatch in mean flux we saw before between PSF and PSF-NOVA and reduces the chances to get outlier data points. 

TODO:

- [x]  Fix `test_perturbation_matrix3d()`

Requires #60 to be merged first.


jorgemarpa:
Using smooth CBVs instead of PCA components and adjusting the priors for the perturbation model improved the deviation from the `mean_model` as seen in the following figures:

x-axis: sum of the `mean_model` in the pixel dimension, i.e. the total PSF per source
y-axis: time-wise mean and standard deviation of the summed `perturbed_model` per source
histogram: ratio of above values

image

Data points of the 1-1 relation are sources with mismatching mean light curves before and after the time model and sources with spiky features (or added variability) after applying the time model.

image

After narrowing the priors the sources with bad time models are fixed.

src/psfmachine/machine.py
```diff
@@ -682,7 +693,7 @@ def build_time_model(
             mpc1_smooth, mpc2_smooth = bspline_smooth(
                 [mpc1, mpc2],
                 x=self.time,
-                do_segments=segments,
+                do_segments=True,
```
christinahedges, line 703: Always?

christinahedges, line 703:
```suggestion
                do_segments=segments,
```

tests/test_perturbation.py
```diff
@@ -86,10 +86,9 @@ def test_perturbation_matrix3d():
     dx, dy = dx.ravel(), dy.ravel()
 
     # ntime x npixels
-    flux = (
-        np.random.normal(0, 0.01, size=(10, 169))
-        + 2 * dx[None, :] ** 2 * 3 * dy[None, :]
-    )
+    flux = np.random.normal(0, 0.01, size=(10, 169)) + dx[None, :] * dy[None, :]
```
christinahedges: Such a nit to pick but you can do 
```suggestion
    flux = np.random.normal(1, 0.01, size=(10, 169)) + dx[None, :] * dy[None, :]
```

christinahedges: And then get rid of line below

christinahedges (commented):
This is 99.999%, I will merge when @jorgemarpa checks that the `segments` keyword is the desired behavior.

src/psfmachine/machine.py
jorgemarpa: I think the smoothing always has to be done by taking care of the segments, otherwise, the smooth version won't reflect the correct behavior of the centroids before/after a break.

tests/test_perturbation.py
jorgemarpa: great!

src/psfmachine/machine.py
jorgemarpa: image

christinahedges: Can we add a log.warning that explains what is happening here

christinahedges (approved):

PR #65, opened 2022-06-30
Normalized Mean Model and PSF metrics

jorgemarpa:
This PR adds two new methods: 

1. A hidden function to find the scale factor that normalized the mean model to sum 1 when sources are full in the TPF.  It creates a high-resolution mean model and integrates the PSF using the trapezoidal rule, the integral is the rescale factor that is applied to the PSF weights. The method also has a diagnostic plot. image

2. A visible method to compute PSF metrics such as:
      - `source_psf_fraction` amount of PSF in data, useful to determine if a source has total or partial flux data. image
      -  `perturbed_ratio_mean` how much the perturbed model deviates from the mean model for a given source, useful to find when the time model changes the mean value of the light curve. The expected value is 1.
      - `perturbed_std` how much variability has the perturbed model in time, helps find when the time model introduces synthetic variability or features. A large value means a poor time model. image

Requires #63 to be merged first


jorgemarpa:
Sanity check, comparing SAP and PSF flux:
before (no normalization)
image

after (normalized shape model)
image


Scatter is due to saturated sources, sources with partial data (e.g. outside or near the edge of the TPF), and highly blended sources.

src/psfmachine/machine.py
```diff
@@ -1127,8 +1129,8 @@ def _update_source_mask_remove_bkg_pixels(self, flux_cut_off=1, frame_index="mea
         #    (self.mean_model.max(axis=1) < 1)
         # )
 
-        # Recreate mean model!
-        self._get_mean_model()
+        # Recreate normalized mean model!
+        self._get_normalized_mean_model()
```
christinahedges, line 1135: Should this be `_renormalize_mean_model()`?

christinahedges: Is there a check to make sure the model isn't already normalized?

src/psfmachine/machine.py
```diff
@@ -1150,6 +1152,185 @@ def _get_mean_model(self):
...
+            cut_r=self.cut_r,
+            n_r_knots=self.n_r_knots,
+            n_phi_knots=self.n_phi_knots,
+        )
+        # evaluate the high res model
+        mean_model_hd = Ap.dot(self.psf_w)
+        mean_model_hd[~np.isfinite(mean_model_hd)] = np.nan
+        mean_model_hd = mean_model_hd.reshape(phi_hd.shape)
+
+        # mask = (mean_model_hd >= -3) & (r_hd < 14)
```
christinahedges: Probably don't need these comments?

src/psfmachine/machine.py
```diff
@@ -1150,6 +1152,185 @@ def _get_mean_model(self):
...
+        a better estimate of `source_psf_fraction`.
+
+        Parameters
+        ----------
+        resolution : int
+            Value in which each pixel axis is split to increase resolution. Default is
+            0 for no subpixel resolution.
+
+        """
+        if resolution > 0:
```
christinahedges: I think something can't have a "resolution" of zero. You might mean `npoints_per_pixel`? I think this is probably ok

christinahedges: How long, roughly, does it take to calculate the high resolution metrics? Should we invest in doing this when we run on the supercomputer?

christinahedges (commented):
This looks very good! Let's add some tests that just check the metrics work and produce reasonable values? Especially the high resolution one would be good to test?

src/psfmachine/machine.py
jorgemarpa: It takes 5 sec to run for ~500 sources with 5400 pixels. 
I think `npoints_per_pixel` is a good name for the argument

src/psfmachine/machine.py
jorgemarpa: Yes, lines`1193-1195` check if the model is normalized. 
The method computes the factor before creating the final `mean_model`, that's why it is named like that.

christinahedges:
@jorgemarpa I reverted these changes because it broke in ways that were hard to debug quickly, you'll need to reopen this pull request!

PR #66, opened 2022-07-01
non-neg pixel flux option and better xknotspacing

jorgemarpa:
This PR improves the knot spacing selection for the background model and adds the option to have positive flux values or zero-centered values for the background.

1. `xknotspacing=6` is a good value when the set of pixels is 'dense' in the row axis but can produce wrong bkg models when it is not as shown here: 
image
which shows significant gaps between bkg pixel data.
image
this is a random selection of bkg pixels

- The solution is to make `xknotspacing` aware of the maximum gap between bkg pixels in the row direction. Here the same example but using half of the maximum gap (16)
image
image

Of course, this sacrifices resolution in the bkg model but should be a good compromise for these cases.

2. An optional argument to make the bkg subtracted flux non-negative, but `machine` in general assumes that bkg pixels are zero centered, therefore using this argument could lead to wrong shape models. Default is set to zero-centered values.

### DIFF
--- a/src/psfmachine/tpf.py
+++ b/src/psfmachine/tpf.py
@@ -200,10 +200,8 @@ class TPFMachine(Machine):
             self.bkg_flux,
             mask=self.bkg_pixel_mask,
             tknotspacing=4,
-            # this should take into account the maximum space between bkg pixels in row
-            xknotspacing=np.maximum(
-                np.diff(np.unique(self.bkg_row[self.bkg_pixel_mask])).max() / 2, 6
-            ),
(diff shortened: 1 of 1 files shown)


christinahedges:
After chatting with @jorgemarpa, we're going to run a few more checks, and find a good number for this. We intend to fix this value for all light curves, so that all the noise properties are as similar as possible across batches!

christinahedges (approved):
This LGTM!

christinahedges:
Great work @jorgemarpa fixing this, I'm happy with the fix!

PR #67, opened 2022-08-05
Adds a limiting magnitude for sources to be considered as contaminants

jorgemarpa:
This PR modifies the following method:

- `machine._get_uncontaminated_sources()` to include a limit magnitude at which sources are considered as contaminants. 
- this limiting magnitude is 17.5 for Kepler data and it is applied when working with crowded regions defined as `n_sources / n_pixels > 0.2`.

### DIFF
--- a/src/psfmachine/machine.py
+++ b/src/psfmachine/machine.py
@@ -139,6 +139,8 @@ class Machine(object):
             the design matrix, options are "linear" or "sqrt".
         quiet: booleans
             Quiets TQDM progress bars.
+        contaminant_mag_limit: float
+          The limiting magnitude at which a sources is considered as contaminant
         """
 
         if not isinstance(sources, pd.DataFrame):
@@ -167,7 +169,7 @@ class Machine(object):
         self.cartesian_knot_spacing = "sqrt"
         # disble tqdm prgress bar when running in HPC
         self.quiet = False
-        self.cont_mag_limit = None
+        self.contaminant_mag_limit = None
 
         if time_mask is None:
             self.time_mask = np.ones(len(time), bool)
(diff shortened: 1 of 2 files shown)


src/psfmachine/machine.py
```diff
@@ -167,6 +167,7 @@ def __init__(
         self.cartesian_knot_spacing = "sqrt"
         # disble tqdm prgress bar when running in HPC
         self.quiet = False
+        self.cont_mag_limit = None
```
christinahedges: What would `cont` mean here? Maybe we can come up with a more descriptive name since I think there will be two magnitude limits and I'm worried it will be confusing.

Astronomy would tell us to use `hard_mag_limit` and `soft_mag_limit` or something but I think that's also not great

src/psfmachine/machine.py
```diff
@@ -564,20 +565,30 @@ def _get_uncontaminated_pixel_mask(self):
         """
         creates a mask of shape nsources x npixels where targets are not contaminated.
         This mask is used to select pixels to build the PSF model.
+
+        Parameters
+        ----------
+        mag_limit: float
+          The magnitude value at which a sources is considered as contaminant
```
christinahedges: I think the docstring for the second limit is missing

src/psfmachine/tpf.py
```diff
@@ -96,6 +96,15 @@ def __init__(
         self.time_corrector = time_corrector
         self.cartesian_knot_spacing = cartesian_knot_spacing
         self.bkg_subtracted = bkg_subtracted
+        # limiting mag values to consider contamination for kepler & tess
+        # hardcoded values to work with kepler
+        if self.nsources / self.npixels > 0.2:
```
christinahedges, line 102: Can we have a log warning message here?

src/psfmachine/tpf.py
```diff
@@ -96,6 +96,15 @@ def __init__(
...
+            if self.tpf_meta["mission"][0].lower() == "kepler":
+                self.cont_mag_limit = 17.5
+            elif self.tpf_meta["mission"][0].lower() == "tess":
+                self.cont_mag_limit = 17
```
christinahedges: Will the source density of 0.2 sources per pixel work well for TESS? I am guessing it will.

christinahedges (commented):
This looks like a great fix. I left a few comments that are very easily closed then we can merge!

src/psfmachine/tpf.py
jorgemarpa: that's a good idea

src/psfmachine/tpf.py
jorgemarpa: I'll do a quick test

src/psfmachine/machine.py
jorgemarpa: supposed to mean contaminant, but yes it isn't intuitive. I think just `contaminant_mag_limit`. Though I want to have it as an internal variable, not a method argument, in that case the log warning message is useful. But, it can still be modified if users want to adjust the `uncontaminated_pixel_mask`.

christinahedges: This sounds perfect!

christinahedges (approved):
LGTM!

christinahedges:
This looks great I'm going to merge. Even though there is no test...I think this is a small change!

PR #69, opened 2022-12-08
Update spline to be slightly more robust

christinahedges:
Extracted out of #54, this is just a slightly more robust version of this part of the code.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ source = ["src"]
 
 [tool.poetry.dependencies]
 python = "^3.7.1"
-numpy = "^1.19.4"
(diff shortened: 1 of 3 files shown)


jorgemarpa:
CI is failing because of an error installing `pyvo v1.1`. It seems that the new version v1.4 might fix the issue, but this requires Astropy v5.x, which only requires Python < 3.8

christinahedges (commented):
This LGTM!

jorgemarpa:
@christinahedges shall we merge this then?

jorgemarpa:
I simplify the code in `make_cartesian_A` and `spline1d` to avoid repeated code, the behavior is the same. 

Merging...

PR #71, opened 2022-12-13
Efficient FFI loading 

jorgemarpa:
This PR updates the FFI loading methods to be more memory efficient plus other minor changes.

Changes:
- loading FFIs using `fitsio` and reading only the asked data
- loading FFIs will attempt to preserve cadence quality. Useful when working with TESS FFIs.
- `ffi.load_from_file()` and `tpf.from_TPFS()` both accept the keyword `sources` for custom user input.
- `ffi.flux_2d` is a property useful for plotting functions
- saturated and bright pixels are masked into `machine.source_mask` and `machine.uncontaminated_pixel_mask`
- updated keyword arguments: 
    - `dist_lim` --> `sparse_dist_lim` as `machine` attribute so it can be modified depending on instrument pixel scale (e.g. TESS)
    - `time_corrector` was deprecated and `Perturbation` handles this
    - `machine.sources_flux_column` to designate globally which column from the `source` catalog will be used as flux estimate. This update is necessary for TESS which uses the R band instead of G.
- Gaia query now also saves RP and BP flux columns
- `plot_shape_model()` has flexible radii and color mapping stretch from flux percentiles. This is useful for other instruments (TESS)


src/psfmachine/ffi.py
```diff
@@ -580,17 +592,21 @@ def _mask_pixels(self, pixel_saturation_limit=1.2e5, magnitude_bright_limit=8):
...
+
+        if not hasattr(self, "source_mask"):
+            # include saturated pixels in the source mask and uncontaminated mask
+            self._get_source_mask()
+            self.source_mask = self.source_mask.multiply(self.pixel_mask).tocsr()
+            self.source_mask.eliminate_zeros()
+            self.uncontaminated_source_mask = self.uncontaminated_source_mask.multiply(
+                self.pixel_mask
+            ).tocsr()
+            self.uncontaminated_source_mask.eliminate_zeros()
```
jorgemarpa: test if this holds when `source_mask` attribute exists.

src/psfmachine/machine.py
jorgemarpa, line 1: Adds an init argument to select the Gaia filter column to be used as flux prior.
Changes the `source_flux_estimate` attribute to be used across the class consistently. 
Makes shape model figure robust to TESS and Kepler data.

src/psfmachine/superstamp.py
jorgemarpa, line 1: updates imports

src/psfmachine/tpf.py
jorgemarpa, line 1: Removes unused attributes.
Adds two attributes to control sparse distance and Gaia column flux.
Updates to dr3 and adds sources catalog as input.

tests/test_ffimachine.py
```diff
@@ -21,14 +21,14 @@ def test_ffi_from_file():
...
-    assert ffi.row.shape == (33812,)
-    assert ffi.ra.shape == (33812,)
-    assert ffi.dec.shape == (33812,)
-    assert ffi.sources.shape == (259, 13)
+    assert ffi.flux_err.shape == (1, 33840)
+    assert ffi.column.shape == (33840,)
+    assert ffi.row.shape == (33840,)
+    assert ffi.ra.shape == (33840,)
+    assert ffi.dec.shape == (33840,)
+    assert ffi.sources.shape == (259, 15)
```
jorgemarpa, line 31: check why change in N pixels

src/psfmachine/utils.py
jorgemarpa, line 1: Adds RP and BP to Gaia query.
Moved `load_FFI_image` and `do_image_cutout` here.

src/psfmachine/ffi.py
jorgemarpa: this is ok. if no source mask exists, then we build one and we have to incorporate saturated and halo pixels. This mask should never be rebuilt after with just `_get_source_mask()` otherwise sat/halo pixels will not be included.

tests/test_ffimachine.py
jorgemarpa: this changed because saturated pixels are treated now. Before they were removed before initializing `machine`, now they still are in the `self.flux` array but they are masked out in the source masks.

jorgemarpa (commented):
Need to run demo notebooks to check consistent results

src/psfmachine/ffi.py
```diff
@@ -580,20 +613,63 @@ def _mask_pixels(self, pixel_saturation_limit=1.2e5, magnitude_bright_limit=8):
...
+            flux_cut_off=flux_cut_off,
+            frame_index=frame_index,
+            bin_data=bin_data,
+            **kwargs,
+        )
+        # include sat/halo pixels again into source_mask
+        self._remove_bad_pixels_from_source_mask()
+        if plot:
+            return self.plot_shape_model(frame_index=frame_index, bin_data=bin_data)
+
```
jorgemarpa, line 672: These are necessary because in FFI we are including saturated and halo pixels in the `source_mask` and `uncontaminated_pixel_mask`. Then if we want to rebuild the source mask using `machine._get_source_mask()` with different params, we need to include the bad pixels again. Same applies for `machine.build_shape_model()` because the method computes a new source_mask in each call in order to account for possible new `flux_cut_off` value (controls how deep in the background the mask goes), we need to add the bad pixels to the new `source_mask`.

src/psfmachine/ffi.py
```diff
@@ -564,6 +585,18 @@ def _bright_sources_mask(self, magnitude_limit=8, tolerance=30):
...
+        Combines source_mask and uncontaminated_pixel_mask with saturated and bright
+        pixel mask.
+        """
+        self.source_mask = self.source_mask.multiply(self.pixel_mask).tocsr()
+        self.source_mask.eliminate_zeros()
+        self.uncontaminated_source_mask = self.uncontaminated_source_mask.multiply(
+            self.pixel_mask
+        ).tocsr()
+        self.uncontaminated_source_mask.eliminate_zeros()
+
```
jorgemarpa, line 599: isolate this in a hidden method so we can call it from other methods

src/psfmachine/machine.py
```diff
@@ -957,16 +959,10 @@ def build_shape_model(
...
-
         # Mask of shape nsources x number of pixels, one where flux from a
         # source exists
-        if not hasattr(self, "source_mask"):
-            self._get_source_mask(**kwargs)
-        # Mask of shape npixels (maybe by nt) where not saturated, not faint,
-        # not contaminated etc
-        self._get_uncontaminated_pixel_mask()
+        # if not hasattr(self, "source_mask"):
+        self._get_source_mask(**kwargs)
```
jorgemarpa, line 965: we need to compute the source masks again in order to account for a possible new `flux_cut_off` value that controls how deep the mask foes into the background level.

jorgemarpa (commented):
lgtm

jorgemarpa (commented):
flake8 passing

jorgemarpa (commented):
flake8 passing

PR #73, opened 2022-12-13
K2 no centroid smoothing

jorgemarpa:
Small fix to avoid smoothing of centroid/poscorr vectors when working with K2 data.

K2 example:
image


christinahedges (commented):
LGTM!

PR #74, opened 2022-12-13
Astropy 5.2 

jorgemarpa:
Updating package dependecies:

- Astropy v5.2
- Python > = 3.8
- kbackground > 0.1.9

This fixes the CI error when installing `pyvo` (an Astropy module) and the error with the units module when using Astropy v4.x with Numpy < 1.22.


christinahedges:
Hi @jorgemarpa this looks great, just need to update the version number in the tests and init file (probably tests/test_psfmachine.py and src/psfmachine/__init__.py) then I'll merge this!

